PR #47, opened 2023-10-19
Automatically update Docker image

### LINKED ISSUE #43: Automatically update Docker image
This could be with each PR or perhaps weekly?

Follows from #11.

joewallwork:
Closes #43.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,11 +1,8 @@
 name: 'Build bespoke Firedrake Docker container'
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-  # TODO: Make it a cron job rather than running in PR testing
+  schedule:
+    - cron: '0 1 * * 6'
 
 jobs:
   docker:


.github/workflows/build.yml
```diff
@@ -1,47 +1,56 @@
...
+
+      - name: 'Lint'
+        if: ${{ always() }}
         run: |
           . /home/firedrake/firedrake/bin/activate
-          python -m pip uninstall -y animate
-          python -m pip install -r requirements.txt
-          python -m pip install -e .
-      - name: Test Animate
+          flake8 --ignore=E501,E226,E402,E741,F403,F405,W503
```
stephankramer, line 41: Are these exceptions not .flake8 already?
It would be annoying if the ones specified here got out of sync with the .flake8
(I know  this was there before)

stephankramer:
Could you say a bit more what you've done here. I can see it builds a firedrake container (docker/Dockerfile.firedrake) - but when does it do this - and does the rebuild of the container trigger a repeat of the test suite?

.github/workflows/build.yml
jwallwork23: Good point, will use the `.flake8` file you introduced in the CI, too.

jwallwork23:
> Could you say a bit more what you've done here. I can see it builds a firedrake container (docker/Dockerfile.firedrake) - but when does it do this - and does the rebuild of the container trigger a repeat of the test suite?

I'm still trying to figure out how to get this working. I'm somewhat of a novice regarding GitHub Actions. If you or @acse-ej321 have any insight I'd be very grateful! I *think* the current attempt tries to build the Docker container upon pushing to the PR. I guess we actually want it to happen upon merging the PR? Or as a last step before merging? Open to ideas.

jwallwork23:
^Various guesses at how to get this working, no joy so far. Will have to read more into how this stuff works.

stephankramer:
Same here - being somewhat of a novice with github actions - normally I just look at an existing project like firedrake and copy that. I guess this setup would be quite a bit different. The Firedrake CI builds and then pushes upon success. What we call "build" here is just running the animate regression tests which should actually be run after we've rebuilt the container i.e. the "docker" target should not depend on "build" but the other way around? The other thing I noticed is that you're referring to a docker/Dockerfile.parmmg (currently commented out) but the repo only has docker/Dockerfile.firedrake

As to when we should rebuild the container: hm not entirely sure. I think regardless of whether we have it trigger on any animate repo activity, we should do a regular rebuild (followed by running animate tests) on a fixed interval, say every week, so that if things break due to changes in the firedrake stack we still catch it if there isn't any recent activity on animate. Then additionally we could also trigger it on PRs/merges into master although maybe that doesn't actually add much (and slows down the acceptance testing) - I guess the only added benefit would be that you would catch if a test/new functionality works in this week's build but not in the very latest but I'm not sure how often that would happen and the number of merges in a week is usually fairly small anyway.

jwallwork23:
Yes that's a good point - I think we'd previously discussed making it a regular build rather than on PR. Until we figure out how to actually do the build, I'll leave it as part of the PR workflow, though.

jwallwork23:
[Apologies if you're getting emails for these commits, it must be annoying!]

Good news is I finally managed to get it to log into DockerHub and the latest workflow run is building!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,36 @@
+name: 'Build bespoke Firedrake Docker container'
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  # TODO: Make it a cron job rather than running in PR testing
```
jwallwork23, line 8: Once the PR is approved, I'll switch the trigger to a cron job before we merge.

stephankramer (approved):
Nice! Thanks for figuring this out and good to know everything still works correctly with latest install

jwallwork23:
@stephankramer Are you happy with my final commit here to set up a cron job for regular testing? https://github.com/pyroteus/animate/pull/47/commits/2b74fcde2c9c51c035d5f2ea535cf9d04fde6b34

stephankramer (approved):
So every Saturday at 1:00, right? Perfect looks good.
I think my other comment was about some commented-out line that has now been removed, yes

PR #58, opened 2023-12-04
First animate demo

stephankramer:
Shows how to setup a simple user-defined metric and adapt the mesh


jwallwork23 (changes requested):
Looks great, thanks very much! A couple of requests and a request suggested by @acse-ej321: how about having each mesh in the right-hand subfigure of a figure with two plots. The left-hand plot would be the initial mesh (i.e., 10x10, 10x10, 10x10, 100x100). It would also be nice to have descriptions of these in the plot titles.

jwallwork23 (approved):

PR #62, opened 2024-01-03
Demo on combining metrics.

stephankramer:


demos/combining_metrics.py
```diff
@@ -0,0 +1,116 @@
...
+mesh_averaged = adapt(mesh, averaged_metric)
+fig, axes = plt.subplots(figsize=(16, 16))
+triplot(mesh_averaged, axes=axes)
+axes.set_aspect('equal')
+axes.set_title('Mesh based on averaged metric')
+fig.show()
+fig.savefig('combining_averaging.jpg')
+
+#
+# .. figure:: combining_averaging.jpg :figwidth: 90% :align: center
```
jwallwork23, line 113: I think these might need to be on separate lines? The image doesn't appear when I build the html.

jwallwork23 (changes requested):
Thanks very much for this, nice demo! I have a few requests but they are mostly just formatting.

stephankramer:
Thanks, Joe. Addressed all points I think and rebased.

jwallwork23 (approved):

PR #63, opened 2024-01-04
Add demos to testing

### LINKED ISSUE #60: Add demos to testing
Following #58.

jwallwork23:
Closes #60.

### DIFF
--- a/test/test_demos.py
+++ b/test/test_demos.py
@@ -1,0 +1,41 @@
+"""
+Checks that all demo scripts run.
+"""
+import glob
+import os
+import pytest
+import shutil
+import subprocess
+import sys
+
+
+# Set environment variable to specify that tests are being run so that we can
+# cut down the length of the demos
+os.environ["ANIMATE_REGRESSION_TEST"] = "1"
+
+cwd = os.path.abspath(os.path.dirname(__file__))
+demo_dir = os.path.abspath(os.path.join(cwd, "..", "demos"))
+all_demos = glob.glob(os.path.join(demo_dir, "*.py"))
+
+
+@pytest.fixture(params=all_demos, ids=lambda x: os.path.basename(x))
+def demo_file(request):
+    return os.path.abspath(request.param)
+
+
+def test_demos(demo_file, tmpdir, monkeypatch):
+    assert os.path.isfile(demo_file), f"Demo file '{demo_file}' not found."
+    demo_name = os.path.splitext(os.path.basename(demo_file))[0]
+
+    # Copy mesh files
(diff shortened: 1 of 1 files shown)


test/test_demos.py
```diff
@@ -0,0 +1,41 @@
...
+# Set environment variable to specify that tests are being run so that we can
+# cut down the length of the demos
+os.environ["ANIMATE_REGRESSION_TEST"] = "1"
+
+cwd = os.path.abspath(os.path.dirname(__file__))
+demo_dir = os.path.abspath(os.path.join(cwd, "..", "demos"))
+all_demos = glob.glob(os.path.join(demo_dir, "*.py"))
+
+
+@pytest.fixture(params=all_demos, ids=lambda x: os.path.basename(x))
```
stephankramer, line 21: The way things are set up here seems to want to future proof for a situation where we have multiple subdirectories under demos/ but in that case I foresee you would not necessarily have unique, meaningful (base) filename (this is currently an issue in Thetis)

test/test_demos.py
```diff
@@ -0,0 +1,41 @@
...
+    for f in glob.glob(os.path.join(source, "*.msh")):
+        shutil.copy(f, str(tmpdir))
+
+    # Change working directory to temporary directory
+    monkeypatch.chdir(tmpdir)
+    subprocess.check_call([sys.executable, demo_file])
+
+    # Clean up plots
+    for ext in ("jpg", "pvd", "vtu"):
+        subprocess.check_call(["rm", "-f", f"{demo_name}*.{ext}"])
```
stephankramer, line 41: So you're assuming a convention where all generated output files are named <basename>.jpg/etc? Sounds sensible but is already not true for the first demo - and isn't strictly followed in goalie either, e.g. burgers_oo.py has burgers-oo_ee.jpg
I'm generally a little wary of duplicating this sort of logic here, that's typically also in the Makefile.

stephankramer (approved):
Happy for this to go in - we can also revisit when we do actually have more complex/multi-directory demos

test/test_demos.py
jwallwork23: To be honest, this was copied from Goalie, which was probably copied from Thetis at some point! Tweaked in https://github.com/pyroteus/animate/pull/63/commits/95b9f3257e2766cfb5aca5d275ceb6af31ab72f5.

test/test_demos.py
jwallwork23: Ah good spot. Fixed in https://github.com/pyroteus/animate/pull/63/commits/6137960d692703273d6eb7295c7421455013915d.

Issue #65, opened 2024-01-08
Set flake8 configuration in setup.cfg


stephankramer:
Adding this in setup.cfg, instead of through command-line options in the Makefile, means flake8 checks do the right thing in IDEs (and in my own vim short-cuts)

stephankramer:
Oh I see you're using .flake8 in goalie. Maybe I should just copy that instead?

jwallwork23:
> Oh I see you're using .flake8 in goalie. Maybe I should just copy that instead?

That'd be great if you wouldn't mind, thanks.

PR #66, opened 2024-01-08
Implements #65

stephankramer:


stephankramer:
Now uses .flake8 instead of setup.cfg to be consistent with goalie

jwallwork23 (approved):
Thanks for this, Stephan!

PR #72, opened 2024-01-26
Option for adapting in serial

### LINKED ISSUE #71: Use subprocess to adapt in serial
ParMmg doesn't support parallel adaptation in 2D. However, solving 2Ds PDEs in parallel is a standard activity of Firedrake. We should use Python `subprocess` to enable serial adaptation during a parallel run.

jwallwork23:
Closes #71.


animate/__init__.py
```diff
@@ -4,3 +4,4 @@
 from animate.metric import *  # noqa
 from animate.recovery import *  # noqa
 from animate.quality import *  # noqa
+from animate.utility import *  # noqa
```
stephankramer, line 7: I think you forgot to commit this

jwallwork23:
Good spot, thanks.

ddundo:
One worrying thing is that checkpointing is very very slow. For example, I ran the burgers demo in serial with two subintervals and checkpointed the final solution of each subinterval - so I'd say it's representative of what this PR is doing. I am attaching the flame graph here.

burgers2

So it looks like checkpointing is more than twice as expensive as `solve_forward`. And similarly for my glacier experiment where I have 100 subintervals (each with 50 timesteps). I also ran the burgers demo with a single subinterval in both serial and parallel (possible since it only has 1 subinterval so there are no mesh-to-mesh projections happening). There is no speed-up in checkpointing, but there is in `solve_forward`.

So if we have to checkpoint at every iteration (we'd have to checkpoint twice in the first iteration and once in latter ones), this might cancel out a lot of the benefits of solving in parallel.

Have you maybe tried this @jwallwork23 or @stephankramer? Any chance there is an alternative?

jwallwork23:
To be honest, I haven't done these kinds of investigations yet so this is useful information, thanks @ddundo.

ddundo:
Thanks @jwallwork23! Yup, here's an example of how I did it

```python
from firedrake import *
from animate.metric import RiemannianMetric
import h5py

# define mesh, metric=RiemannianMetric(...)

mp = {
    "dm_plex_metric_target_complexity": 1234.,
    "dm_plex_metric_h_min": 1.,
    ...
}

# checkpoint metric
with CheckpointFile('test.h5', 'w') as afile:
    afile.save_function(metric, name="metric")

# save parameters under path metric_parameters/{parameter_name}
with h5py.File('test.h5', 'a') as h5file:
    for key, value in mp.items():
        h5file.create_dataset(f'metric_parameters/{key}', data=value)

# and then load like this
with h5py.File('test.h5', 'r') as h5file:
    loaded_mp = {}
    for key in mp.keys():
        loaded_mp[key] = np.array(h5file[f'metric_parameters/{key}'])
```

jwallwork23:
This works locally but annoyingly the GitHub Actions run gets stuck on the first test that uses subprocess. I wonder if it has a problem with it.

stephankramer:
Out of interest why do we need subprocess to do this? i.e. why don't we just do the whole "read checkpoint, adapt, write checkpoint" in the same process on rank 0?

ddundo:
> Out of interest why do we need subprocess to do this? i.e. why don't we just do the whole "read checkpoint, adapt, write checkpoint" in the same process on rank 0?

Joe mentioned subprocess before I [commented this](https://github.com/pyroteus/goalie/issues/96#issuecomment-1910149703), but I now went to quantify it and realised I was loading it wrong in the case that Stephan is referring to. So maybe it's not slower, but it is a lot more complex (and maybe impossible?)

- So with subprocess we can do:
```python
if rank == 0:
    subprocess.run(["mpirun", "-n", "1", "python", "adapt_script.py"])
COMM_WORLD.barrier()

##### in adapt_script.py:
with CheckpointFile('save_old.h5', 'r') as afile:
    loaded_mesh = afile.load_mesh("firedrake_default")
    loaded_metric = afile.load_function(loaded_mesh, "metric")
    loaded_mp = afile._read_pickled_dict("metric_parameters", "mp_dict")
        
loaded_RM = RiemannianMetric(loaded_metric.function_space()).assign(loaded_metric)
loaded_RM.set_parameters(loaded_mp)
adapted_mesh = adapt(loaded_mesh, loaded_RM)

with CheckpointFile('save_new.h5', 'w') as afile:
    afile.save_mesh(adapted_mesh)
```
- But in Stephan's method it's a lot more complex
```python
if rank == 0:
    with CheckpointFile('save_old.h5', 'r', comm=COMM_SELF) as afile:
        loaded_mesh = afile.load_mesh("firedrake_default")
        loaded_metric = afile.load_function(loaded_mesh, "metric")
        loaded_mp = afile._read_pickled_dict("metric_parameters", "mp_dict")

    new_metric.set_parameters(loaded_mp)
    adapted_mesh = adapt(new_mesh, new_metric)

    with CheckpointFile('save_new.h5', 'w', comm=COMM_SELF) as bfile:
        bfile.save_mesh(adapted_mesh)
COMM_WORLD.barrier()
```
We have to pass `COMM_SELF` to `CheckpointFile` or it will never load. But even here it gives this error:
 ```
   File "/data2/testing/par/nosubpr_burgers1.py", line 210, in <module>
    solutions = mesh_seq.fixed_point_iteration(adaptor)
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/ubuntu/software/firedrake-feb24/src/goalie/goalie/mesh_seq.py", line 677, in fixed_point_iteration
    continue_unconditionally = adaptor(self, self.solutions, **adaptor_kwargs)
  File "/data2/testing/par/nosubpr_burgers1.py", line 185, in adaptor
    adapted_mesh = adapt(new_mesh, new_metric)
  File "/home/ubuntu/software/firedrake-feb24/src/animate/animate/adapt.py", line 137, in adapt
    return adaptor.adapted_mesh
  File "/home/ubuntu/software/firedrake-feb24/src/PyOP2/pyop2/utils.py", line 62, in __get__
    obj.__dict__[self.__name__] = result = self.fget(obj)
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/ubuntu/software/firedrake-feb24/src/animate/animate/adapt.py", line 84, in adapted_mesh
    return fmesh.Mesh(
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/ubuntu/software/firedrake-feb24/src/firedrake/firedrake/mesh.py", line 2690, in Mesh
    raise ValueError("Communicator used to create `plex` must be at least congruent to the communicator used to create the mesh")
ValueError: Communicator used to create `plex` must be at least congruent to the communicator used to create the mesh
application called MPI_Abort(PYOP2_COMM_WORLD, 1) - process 0
```

So I tried defining a new mesh that is based on the loaded mesh and passing `COMM_SELF` to it: `new_mesh = Mesh(loaded_mesh, comm=COMM_SELF)` but `new_mesh.topology_dm.comm` is unchanged. And I see that `CheckpointFile._load_mesh_topology` does seem to consider the `comm` that is passed... So I don't know!

**Comment by ddundo** [comment]
@jwallwork23 if you'd rather not use private methods, we can also do

```
with CheckpointFile('test.h5', 'w') as afile:
    afile.save_function(metric, name="metric")
    for key, value in mp.items():
        afile.h5pyfile.create_dataset(f'metric_parameters/{key}', data=value)
```

(Essentially the same as the first way I suggested and then edited which used `h5py` but all in one)

**Comment by jwallwork23** [comment]
Thanks both! That works locally and simplifies things a lot! Will tidy things up. Downside is that the CI still seems to hang, will need to investigate.

**Comment by jwallwork23** [comment]
Now tidied up and fixed some oversights in the checkpointing. Ready for review, although the CI is still hanging as soon as it hits a test with the new functionality. Any thoughts?

File: animate/adapt.py [hunk 2]
```diff
@@ -111,27 +138,84 @@ def interpolate(self, f):
...
+
+        # In parallel, save input mesh and metric to a checkpoint file
+        save_checkpoint(metric, metric_fname, metric_name=metric_name)
+
+        # In serial, load the checkpoint, adapt and write out the result
+        if COMM_WORLD.rank == 0:
+            metric0 = load_checkpoint(metric_fname, metric_name, comm=COMM_SELF)
+            mesh0 = metric0._mesh
+            adaptor0 = MetricBasedAdaptor(
+                mesh0, metric0, name=mesh_name, comm=COMM_SELF
```
ddundo, line 202: The `name` argument should be `name=name`, i.e. the one that the user passed to `adapt`. This is how you do it in L184 too.

animate/adapt.py
```diff
@@ -41,11 +56,16 @@ class MetricBasedAdaptor(AdaptorBase):
     """
 
     @PETSc.Log.EventDecorator()
-    def __init__(self, mesh, metric, name=None):
+    def __init__(self, mesh, metric, name=None, comm=COMM_WORLD):
```
ddundo, line 59: *** Marking all connected comments with these stars ***

`comm=None`

animate/adapt.py
```diff
@@ -61,14 +81,16 @@ def __init__(self, mesh, metric, name=None):
         if name is None:
             name = mesh.name
         self.name = name
+        self.comm = comm
```
ddundo, line 84: *** Marking all connected comments with these stars ***

Do the same as for `name`:
```python
if commis None:
    comm= mesh.comm
self.comm = comm
```

`mesh.comm` will be `COMM_SELF` because we loaded it by calling `CheckpointFile(..., comm=COMM_SELF)`

animate/adapt.py
ddundo, line 202: *** Marking all connected comments with these stars ***

I guess you separated `mesh0` so that `adaptor0` would fit in a single line, but it still doesn't. In fact, it's one character too long! But if you change the `name` argument to `name=name` and don't pass `comm=COMM_SELF` (see *** comments), then this fits into one line: `adaptor0 = MetricBasedAdaptor(metric0_mesh, metric0, name=name)`

animate/adapt.py
ddundo, line 84: *** Marking all connected comments with these stars ***

We could pass both `name` and `comm` to `super().__init__(mesh, name=name, comm=comm)` since that is where `self.mesh` is defined?

animate/adapt.py
```diff
@@ -111,27 +138,84 @@ def interpolate(self, f):
...
     if len(metrics) > 1:
         metric.intersect(*metrics[1:])
+
     adaptor = MetricBasedAdaptor(mesh, metric, name=name)
-    return adaptor.adapted_mesh
+    if serialise:
+        checkpoint_dir = get_checkpoint_dir()
+        if not os.path.exists(checkpoint_dir):
+            os.makedirs(checkpoint_dir)
+        metric_name, mesh_name = "tmp_metric", "tmp_adapted_mesh"
```
ddundo, line 189: --- Marking all connected comments with these dashes---

`mesh_name` here is the name of the mesh that the user defined in their script, so should be `mesh_name = mesh.name`

animate/adapt.py
```diff
@@ -111,27 +138,84 @@ def interpolate(self, f):
...
+        metric_fname = "metric_checkpoint"
+        input_fname = os.path.join(checkpoint_dir, metric_fname + ".h5")
+        output_fname = os.path.join(checkpoint_dir, "adapted_mesh_checkpoint.h5")
+
+        # In parallel, save input mesh and metric to a checkpoint file
+        save_checkpoint(metric, metric_fname, metric_name=metric_name)
+
+        # In serial, load the checkpoint, adapt and write out the result
+        if COMM_WORLD.rank == 0:
+            metric0 = load_checkpoint(metric_fname, metric_name, comm=COMM_SELF)
```
ddundo, line 199: --- Marking all connected comments with these dashes---

After the change in L189, we need to pass `mesh_name` as an argument here:
`metric0 = load_checkpoint(metric_fname, mesh_name, metric_name, comm=COMM_SELF)` (see comment in load_checkpoint)

This doesn't fit in a single line so might use some renaming :)

animate/checkpointing.py
```diff
@@ -0,0 +1,97 @@
...
+    :kwarg comm: MPI communicator for handling the checkpoint file
+    :type comm: :class:`mpi4py.MPI.Intracom`
+    :returns: the metric loaded from the checkpoint
+    :rtype: :class:`animate.metric.RiemannianMetric`
+    """
+    fname = _fix_checkpoint_filename(filename)
+    if not os.path.exists(fname):
+        raise Exception(f"Metric file does not exist! Path: {fname}.")
+    with fchk.CheckpointFile(fname, "r", comm=comm) as chk:
+        mesh = chk.load_mesh()
```
ddundo, line 52: --- Marking all connected comments with these dashes---

L32 should be `def load_checkpoint(filename, mesh_name, metric_name, comm=firedrake.COMM_WORLD):`

and L52 `mesh = chk.load_mesh(mesh_name)`

ddundo (changes requested):
Thanks a lot for doing this @jwallwork23! I left some comments :)

And sorry, I have no idea about the CI question.

ddundo:
Regarding my flamegraph above... you made me realise what was wrong @jwallwork23! I am doing stuff on a server where I have the firedrake installation and my data on separate drives, and they apparently have very different writing speeds. So when I checkpoint into the drive with the firedrake installation (as your PR here does), it takes less than a second, and on the other drive it takes over 30 seconds.

So you might want to check how long checkpointing takes for you just in case, but I assume that it will not be slow! So ignore my comment above :) thanks for this!

animate/adapt.py
```diff
@@ -111,27 +138,84 @@ def interpolate(self, f):
...
+        :returns: its interpolation onto the adapted mesh
+        :rtype: :class:`firedrake.function.Function`
         """
         raise NotImplementedError(
             "Consistent interpolation has not yet been implemented in parallel"
         )  # TODO
 
 
-def adapt(mesh, *metrics, name=None):
+def adapt(mesh, *metrics, name=None, serialise=False, remove_checkpoints=True):
```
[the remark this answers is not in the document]
jwallwork23: That's not true. If we are running 3D in parallel passing `serialise=True` then it will stay true.

animate/adapt.py
```diff
@@ -111,27 +138,84 @@ def interpolate(self, f):
...
+    :type serialise: :class:`bool`
+    :kwarg remove_checkpoints: if ``True``, checkpoint files are deleted after use
+    :type remove_checkpoints: :class:`bool`
+    :returns: the adapted mesh
+    :rtype: :class:`~firedrake.mesh.MeshGeometry`
     """
+
+    # Parallel adaptation is currently only supported in 3D
+    if mesh.topological_dimension() != 3:
+        serialise = False
```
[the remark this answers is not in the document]
jwallwork23: Why?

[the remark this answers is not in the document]
ddundo: If our mesh is 2D and we do `adapt(mesh, metric, serialise=True)` with
```
if mesh.topological_dimension() != 3:
    serialise = False
```
returns
```
[1] DMAdaptMetric() at /home/ubuntu/software/firedrake-feb24/src/petsc/src/dm/interface/dmgenerate.c:249
[1] DMAdaptMetric_ParMmg_Plex() at /home/ubuntu/software/firedrake-feb24/src/petsc/src/dm/impls/plex/adaptors/parmmg/parmmgadapt.c:61
[1] Argument out of range
[1] ParMmg only works in 3D.
```

animate/adapt.py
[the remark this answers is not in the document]
ddundo: I think there is confusion with the terminology used, which might be the cause of the confusion in https://github.com/pyroteus/animate/pull/72#discussion_r1517656191 too :) why would we pass `serialise=True` if we are running 3D in parallel? In that case, we do NOT serialise since it can work in parallel, right?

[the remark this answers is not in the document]
ddundo: Sorry, I edited the comment while you were writing this I think. Please see the new one :)

animate/adapt.py
```diff
@@ -111,27 +138,84 @@ def interpolate(self, f):
...
     metric = metrics[0]
     if len(metrics) > 1:
         metric.intersect(*metrics[1:])
+
     adaptor = MetricBasedAdaptor(mesh, metric, name=name)
-    return adaptor.adapted_mesh
+    if serialise:
+        checkpoint_dir = get_checkpoint_dir()
+        if not os.path.exists(checkpoint_dir):
+            os.makedirs(checkpoint_dir)
```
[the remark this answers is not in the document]
jwallwork23: Good spot. Fixed in 90facc3.

animate/adapt.py
```diff
@@ -111,27 +138,84 @@ def interpolate(self, f):
...
+        if not os.path.exists(output_fname):
+            raise Exception(f"Adapted mesh file does not exist! Path: {output_fname}.")
+        with fchk.CheckpointFile(output_fname, "r") as chk:
+            newmesh = chk.load_mesh(mesh_name)
+
+        # Delete temporary checkpoint files
+        if remove_checkpoints and COMM_WORLD.rank == 0:
+            os.remove(input_fname)
+            os.remove(output_fname)
+        COMM_WORLD.barrier()
```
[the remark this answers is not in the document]
jwallwork23: I suppose not. Fixed in 90facc3.

animate/adapt.py
```diff
@@ -111,27 +138,84 @@ def interpolate(self, f):
...
+    # If already running in serial then no need to use checkpointing
+    if COMM_WORLD.size == 1:
+        serialise = False
+
+    # Combine metrics by intersection, if multiple are passed
     metric = metrics[0]
     if len(metrics) > 1:
         metric.intersect(*metrics[1:])
+
     adaptor = MetricBasedAdaptor(mesh, metric, name=name)
```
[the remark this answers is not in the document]
jwallwork23: Fixed in 61717de.

animate/adapt.py
jwallwork23: Fixed in 61717de.

jwallwork23: Thanks but I don't think we should change the API just to make characters fit on a line.

animate/adapt.py
jwallwork23: Fixed in dab6246.

animate/adapt.py
jwallwork23: What about the change in 61717de?

animate/adapt.py
jwallwork23: Why? `None` is not a valid communicator.

animate/adapt.py
jwallwork23: Fixed in 83d8b81.

animate/adapt.py
jwallwork23: Fixed in 83d8b81.

animate/adapt.py
jwallwork23: Fixed in 83d8b81.

animate/adapt.py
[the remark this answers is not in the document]
jwallwork23: There's no point serialising if you're already running in serial - as you've noted, it will be more expensive to use the checkpointing approach.

The 3D parallel case with ParMmg is not very well tested and likely has bugs, so I can see cases where it would be useful to serialise and run with Mmg.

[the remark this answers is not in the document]
jwallwork23: We might want to serialise 3D parallel because ParMmg can be somewhat unreliable.

animate/adapt.py
ddundo: Sorry, that doesn't work, no. The `name` that we pass to `new_mesh = adapt(old_mesh, metric, name=name)`) is the name of the adapted mesh: `new_mesh.name`. And `mesh_name` in L189 is `old_mesh.name`, which the user might have set to something when constructing the mesh, e.g. `old_mesh = RectangleMesh(..., name='original_mesh')`.

So when you do 
```
with fchk.CheckpointFile(fname, "r", comm=comm) as chk:
    mesh = chk.load_mesh()
```
in `load_checkpoint`, this assumes by default `mesh = chk.load_mesh('firedrake_default')`, whereas we need to do `mesh = chk.load_mesh('original_mesh')`

animate/adapt.py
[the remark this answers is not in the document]
jwallwork23: Oh I see, good point. Fixed in 6ac1114.

animate/adapt.py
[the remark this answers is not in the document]
jwallwork23: Here you are referring to Goalie, not Animate. Let's not try to get ahead of ourselves in this PR.

animate/adapt.py
jwallwork23: I couldn't get this to work. I won't have chance to work on this for a few days so if you have a working solution please feel free to push to this branch.

animate/checkpointing.py
jwallwork23: I couldn't get this to work. I won't have chance to work on this for a few days so if you have a working solution please feel free to push to this branch.

ddundo: Fixed in 7fcce996cd69fb074e6bf6657c6346631053461f .

Here is the MFE if you want to compare between commits:
```
old_mesh = UnitSquareMesh(3, 3, name="old_mesh")
metric = RiemannianMetric(TensorFunctionSpace(old_mesh, "CG", 1))
metric.set_parameters({})
new_mesh = adapt(old_mesh, metric, name="new_mesh")
```
Gives:
```
Traceback (most recent call last):
  File "/data2/testing/par/animate/mfe.py", line 9, in <module>
    new_mesh = adapt(old_mesh, metric, name="new_mesh")
  File "/home/ubuntu/software/firedrake-feb24/src/animate/animate/adapt.py", line 203, in adapt
    metric0 = load_checkpoint(metric_fname, metric_name, comm=COMM_SELF)
  File "/home/ubuntu/software/firedrake-feb24/src/animate/animate/checkpointing.py", line 52, in load_checkpoint
    mesh = chk.load_mesh()
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/ubuntu/software/firedrake-feb24/src/firedrake/firedrake/checkpointing.py", line 1035, in load_mesh
    tmesh_name = self._get_mesh_name_topology_name_map()[name]
KeyError: 'firedrake_default'
```
because it expects that `old_mesh.name = 'firedrake_default'`.

animate/adapt.py
ddundo: Fixed in 7fcce996cd69fb074e6bf6657c6346631053461f . See https://github.com/pyroteus/animate/pull/72#discussion_r1517913907 for MFE before the fix.

animate/adapt.py
ddundo: Fixed in 7fcce996cd69fb074e6bf6657c6346631053461f . See https://github.com/pyroteus/animate/pull/72#discussion_r1517913907 for MFE before the fix.

animate/adapt.py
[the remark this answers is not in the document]
ddundo: Yeah I was just thinking about when this would be useful :) will mark this as resolved then!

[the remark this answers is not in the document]
ddundo: I still find this very confusing :) So if we know that parallel adaptation only works in 3D, I don't like having the `serialise` argument to `adapt`. Especially since we have the `if` checks in place. 

So if the idea is to adapt in parallel only if it's 3D unless we specifically want to serialise 3D due to ParMmg being unreliable sometimes, how about this:

```
def adapt(mesh, *metrics, name=None, serialise3d=False, remove_checkpoints=True):
    nprocs = COMM_WORLD.size

    # Parallel adaptation is currently only supported in 3D
    if mesh.topological_dimension() != 3:
        serialise = nprocs >= 1  # always serialise if it is not 3D
    elif serialise3d:
        serialise = nprocs > 1  #  serialise 3D if serialise3d AND if we aren't already running in serial (i.e. avoid unnecessary checkpointing)
```

animate/checkpointing.py
```diff
@@ -63,7 +63,7 @@ def load_checkpoint(filename, metric_name, comm=firedrake.COMM_WORLD):
     return metric
 
 
-def save_checkpoint(metric, filename, metric_name=None, comm=firedrake.COMM_WORLD):
+def save_checkpoint(filename, metric, metric_name, comm=firedrake.COMM_WORLD):
```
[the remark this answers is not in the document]
ddundo: The alternative is to do `metric.rename(metric_name)` at some point. Feel free to revert this if you prefer it otherwise @jwallwork23 :)

animate/adapt.py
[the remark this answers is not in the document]
ddundo: Could also raise a warning if `serialise=True`.

animate/checkpointing.py
ddundo: @jwallwork23 if you think this is confusing for future users/developers I can also make some comments and give more descriptive variable names. Happy to do it if you want me to :)

animate/adapt.py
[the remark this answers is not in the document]
ddundo: I just noticed. `get_checkpoint_dir` (line above) already makes the directory.

animate/utility.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    Retrieve the path to Animate's checkpoint directory.
+    """
+    if os.environ.get("ANIMATE_CHECKPOINT_DIR"):
+        checkpoint_dir = os.environ["ANIMATE_CHECKPOINT_DIR"]
+    else:
+        import animate
+
+        assert animate.__file__[-19:] == "animate/__init__.py"
+        checkpoint_dir = os.path.join(animate.__file__[:-20], ".checkpoints")
```
ddundo, line 17: Maybe nicer to do this?

```
assert os.path.basename(animate.__file__) == "__init__.py"
checkpoint_dir = os.path.join(os.path.dirname(animate.__file__), ".checkpoints")
```

jwallwork23:
@ddundo Thanks for these changes, although the tests no long pass locally for me. Do they work for you? If not, please could you fix them?

animate/utility.py
jwallwork23: Thanks. Fixed in 3c0e1ed.

ddundo:
> @ddundo Thanks for these changes, although the tests no long pass locally for me. Do they work for you? If not, please could you fix them?

Sorry, thought you'd review the changes first :) Done in dabb9f68fcb66f72b761c10444410799d77a2268. I added a new unit test for custom mesh names which required me to slightly modify other tests too.

animate/utility.py
jwallwork23: Ah yes.. Fixed in 161a667.

jwallwork23:
I'm afraid tests still fail locally for me.

jwallwork23:
Thanks @ddundo.

Okay my recent debugging commits show that the rank 0 if condition works fine... but then it never leaves the if block. This isn't something I've seen before - let's discuss in the meeting.

ddundo:
@jwallwork23 I got it to work! Please see the successful commit at https://github.com/ddundo/animate/pull/1 I added `time.sleep(30)` for the rank 1 process

jwallwork23:
> @jwallwork23 I got it to work! Please see the successful commit at [ddundo#1](https://github.com/ddundo/animate/pull/1) I added `time.sleep(30)` for the rank 1 process

Nice work! Although this is bizarre...

ddundo:
> Nice work! Although this is bizarre...

My theory is that the CI doesn't like having dead processes so we need to have the rank 1 process do something. Which I guess makes sense!

jwallwork23:
> My theory is that the CI doesn't like having dead processes so we need to have the rank 1 process do something. Which I guess makes sense!

So https://github.com/pyroteus/animate/commit/b97ba84d5aca8ebb81eb15bdfe9274f67b89a23d worked. However, it seems very temperamental because https://github.com/pyroteus/animate/pull/72/commits/47ab149fdb42444260ace14037b2484d9921db1a is indentical but it hangs...

ddundo:
> So [b97ba84](https://github.com/pyroteus/animate/commit/b97ba84d5aca8ebb81eb15bdfe9274f67b89a23d) worked. However, it seems very temperamental because [47ab149](https://github.com/pyroteus/animate/commit/47ab149fdb42444260ace14037b2484d9921db1a) is indentical but it hangs...

Yeah I had a similar issue, where just adding a comment would cause it to hang. Do you think it might be worth opening a ticket for github support? Seems like it's not something from our side

ddundo:
Okay, I'm pretty sure I finally fixed it! When it seems to randomly hang, it's because an error was raised on one process. And this doesn't seem to kill the other ranks if there is a barrier afterwards. For example, try running this MFE with and without the barrier:
```
from mpi4py import MPI

if MPI.COMM_WORLD.rank == 0:
    raise ValueError("Test")
MPI.COMM_WORLD.barrier()  # it hangs forever if there is a barrier
```
I was googling and found only super complicated ways to broadcast an error from the process on which it was raised to all others. Do you or @stephankramer know of a simple way to do it? Otherwise, I think we should just get rid of all barriers, as I did here.

1. First barrier: when it got stuck without printing your debug messages it turns out that it got stuck here:
```
        checkpoint_dir = get_checkpoint_dir()
        if not os.path.exists(checkpoint_dir) and COMM_WORLD.rank == 0:
            os.makedirs(checkpoint_dir)
        COMM_WORLD.barrier()
```
This is probably because of the `FileExistsError` that I mentioned in this https://github.com/pyroteus/animate/pull/72#discussion_r1517671058. I don't know if you saw my comment afterwards on the same thread, but `checkpoint_dir` already gets created in `get_checkpoint_dir`. So I would remove this if condition completely and inside `get_checkpoint_dir` change `os.makedirs(checkpoint_dir)` to `os.makedirs(checkpoint_dir, exist_ok=True)`. This was why it was "temperamental". 

2. Second barrier: I replaced the barrier with something else which acts like a barrier, using `time.sleep` and `COMM_WORLD.bcast`.

jwallwork23:
Nice work @ddundo! Thanks very much for this. I added a minor refactoring to avoid sleeping unnecessarily long.

This is now finally ready for review. Would you mind taking a look @stephankramer?

stephankramer (changes requested):

stephankramer:
Ah sorry, I thought I had already left some comments - but these were still "pending". I think my point re. the checkpoint file naming may be relevant for the discussion above - I don't know if we are running any tests concurrently?

Re: the barrier. I'm afraid it's not really a solution to just not have barriers. The issue in the MFE you gave is that the error you raise does not call MPI.Abort - thus the process exits quietly leaving the other one hanging (correctly). This is prevented in firedrake world by adding an exception hook (see PyOP2/mpi.py). In general, it's just a fact of life that in parallel you won't always get complete output, hangs when error occurs.

I don't think calling makedirs with exists_ok=True is safe - you may prevent an error but you're still creating a race condition with multiple processes trying to create the same directory at the same time. The thing is, under the hood the making of a process still consists of several steps that the several processes go through and it is very well possible that at the point one process tries to detect whether the path already exists, another process has already started its process of creating that same path but without making any detectable changes yet. E.g. this is *not* safe in parallel
```
if not path.exists(PATH):
   makedir(PATH)
```
because multiple processes may pass the tests simultaneously.

Similarly replacing a barrier with sleep+bcast feels like voodoo programming to me (a common issue with people trying to debug MPI issues). The point of barriers is actually to make a parallel program more predictable. So if it doesn't work with a barrier there's probably something wrong...

ddundo:
Thanks @stephankramer! I should say re. the second barrier I removed: I haven't run into errors there so leaving the second barrier works. I just removed it "in case" an error would be raised for whatever reason - but these are harder to predict than the `FileExistsError` in the first barrier (e.g. `save_checkpoint` might fail because the user has no space left, etc.).

ddundo:
@jwallwork23 and @stephankramer, how about then adding a try and except block? This would allow us to keep the barrier and still kill the program, but we wouldn't know why (please improve the idea if you think it has merit!):
```
saved = False
if COMM_WORLD.rank == 0:
    try:
        # checkpoint and adapt...
        saved = True
    except:
        raise ValueError("Something failed.")
COMM_WORLD.barrier()
if not saved:
    raise ValueError("Something failed.")
```

stephankramer:
So to be clear, and also make sure I'm not misunderstanding things.

So I thought the issue you described was caused by creating the directory in two different places which is now fixed.

Then the reason this exception caused a hang is still unclear. Whenever we're in animate we have already import firedrake, which in turn imports pyop2 which should have added the exception hook that causes it to mpi.abort. Maybe this doesn't happen because the exception is caught by pytest first. Actually let me test that.....

Yes, if you make the following test:
```
import firedrake
from mpi4py import MPI 

@pytest.mark.parallel(nprocs=2)
def test_mine():

    if MPI.COMM_WORLD.rank == 0:
        raise ValueError("Test")
    MPI.COMM_WORLD.barrier()  # it hangs forever if there is a barrier
``
and run it under pytest, it hangs - whereas if you just run it with mpiexec and call test_mine() it doesn't (because of the firedrake import). So this is something we should probably report to pytest-mpi rather than trying to work around in a specific place. Note that this might happen anywhere when running in parallel, not just in the specific place where we fork of a separate code path on rank 0. Note that this has nothing to do with using barriers or not. Any time you run in parallel, and one of the processes silently exits (without calling abort) you will get a hang because somewhere further on another process will be waiting for a message that the exiting process is meant to have sent.

In any case, we need a barrier right after we create the directory. As we really should only give the makedir call on one rank, but all processes need to write to it initially (before rank 0  then reads it back in). So before they write we need a guarantee that the directory is actually created.

**Comment by ddundo** [comment]
Ah that's interesting! Thanks for explaining and testing this!

File: animate/adapt.py [hunk 15]
```diff
@@ -15,22 +20,38 @@ class AdaptorBase(abc.ABC):
...
 
-    def __init__(self, mesh):
+    def __init__(self, mesh, name=None, comm=None):
         """
-        :param mesh: mesh to be adapted
+        :arg mesh: mesh to be adapted
+        :type mesh: :class:`firedrake.mesh.MeshGeometry`
+        :kwarg name: name for the adapted mesh
+        :type name: :class:`str`
+        :kwarg comm: MPI communicator for handling the checkpoint file
```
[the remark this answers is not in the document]
jwallwork23: Good point. Fixed in 67c1d52da2b40efc963de75c7207a25b39dff965.

I suppose we will probably only consider metric-based approaches so I suppose we could probably drop this ABC, indeed.

animate/adapt.py
```diff
@@ -111,27 +140,83 @@ def interpolate(self, f):
...
         metric.intersect(*metrics[1:])
-    adaptor = MetricBasedAdaptor(mesh, metric, name=name)
-    return adaptor.adapted_mesh
+
+    if serialise:
+        checkpoint_dir = get_checkpoint_dir()
+        if not os.path.exists(checkpoint_dir) and COMM_WORLD.rank == 0:
+            os.makedirs(checkpoint_dir)
+        COMM_WORLD.barrier()
+        metric_name = "tmp_metric"
```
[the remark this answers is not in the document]
jwallwork23: Ah yeah good point. Putting this in hits the hang bug mentioned below.

jwallwork23:
> ```
> import firedrake
> from mpi4py import MPI 
> 
> @pytest.mark.parallel(nprocs=2)
> def test_mine():
> 
>     if MPI.COMM_WORLD.rank == 0:
>         raise ValueError("Test")
>     MPI.COMM_WORLD.barrier()  # it hangs forever if there is a barrier
> ```

Thanks very much for looking into this in detail @stephankramer. So just to be clear, your recommendation is to post this as an issue on the `pytest-mpi` repo?

ddundo:
This hangs again... I'll investigate

jwallwork23:
> This hangs again... I'll investigate

Alas... but thanks!

ddundo:
Okay this is ready now! Sorry for pushing to your branch - it was a minor fix :)

But it was still annoying to debug since it hangs and we don't get an error message. So I'd be keen to report [this](https://github.com/mesh-adaptation/animate/pull/72#issuecomment-2000169754) to pytest-mpi to avoid issues in the future, but I think it would be better if you or @stephankramer did it :)

ddundo (approved):

jwallwork23:
Okay cool, thanks @ddundo. Will leave it to @stephankramer to re-review then.

stephankramer (approved):
Looks good. Some minor suggestions

ddundo:
Thanks for reviewing @stephankramer! 

@jwallwork23 could you please take a look and see if you agree, and then I can make the final changes?

jwallwork23:
> Thanks for reviewing @stephankramer!
> 
> @jwallwork23 could you please take a look and see if you agree, and then I can make the final changes?

That'd be great, thanks @ddundo.

animate/adapt.py
```diff
@@ -113,27 +142,73 @@ def interpolate(self, f):
...
-    :param metrics: list of :class:`.RiemannianMetric`\s
-    :param name: name for the adapted mesh
-    :return: a new :class:`~firedrake.mesh.MeshGeometry`.
+    :arg mesh: mesh to be adapted.
+    :type mesh: :class:`firedrake.mesh.MeshGeometry`
+    :arg metrics: metrics to guide the mesh adaptation
+    :type metrics: :class:`list` of :class:`.RiemannianMetric`\s
+    :kwarg name: name for the adapted mesh
+    :type name: :class:`str`
+    :kwarg serialise: if ``True``, adaptation is done in serial
```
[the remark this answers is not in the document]
ddundo: Added in 6e1fc55a89ebd6a1af154e4ee5fbcc85a3851dda

ddundo:
@jwallwork23 I addressed everything now and added a `tearDownClass` method in `test_checkpoint.py` to delete all temporary directories created during testing there. I'm happy to merge if you are :)

animate/adapt.py
```diff
@@ -173,13 +176,12 @@ def adapt(mesh, *metrics, name=None, serialise=False, remove_checkpoints=True):
...
     nprocs = COMM_WORLD.size
 
-    # Parallel adaptation is currently only supported in 3D
-    if mesh.topological_dimension() != 3:
-        serialise = nprocs > 1
-
-    # If already running in serial then no need to use checkpointing
-    if nprocs == 1:
-        serialise = False
+    PETSc.Sys.Print(f"{serialise=}", comm=COMM_SELF)
```
jwallwork23: Wow, I've never seen this notation before!

jwallwork23:
> @jwallwork23 I addressed everything now and added a `tearDownClass` method in `test_checkpoint.py` to delete all temporary directories created during testing there. I'm happy to merge if you are :)

Excellent! Merging now.

animate/adapt.py
ddundo: Very useful for debugging! That's why I forgot to remove it here, but did in a later commit :)

ddundo:
@jwallwork23 I just realised. You opened https://github.com/aragilar/pytest-mpi/issues/49 but I think it should be reported to Firedrake's https://github.com/firedrakeproject/mpi-pytest?

jwallwork23:
> @jwallwork23 I just realised. You opened [aragilar/pytest-mpi#49](https://github.com/aragilar/pytest-mpi/issues/49) but I think it should be reported to Firedrake's https://github.com/firedrakeproject/mpi-pytest?

Oh okay, thanks. Will close that one and open one there.

Issue #73, opened 2024-01-27
ImportError calling triplot as function has moved in firedrake


acse-ej321:
Both animate demos currently throw ImportErrors following the migration of `triplot` in firedrake
Calling:
https://github.com/pyroteus/animate/blob/78bc2772cf41ccc502d4c6f577c635ad9dd066e8/demos/simple_metric.py#L51
Results in:

ImportError: The function `triplot` has moved to `firedrake.pyplot`, update your code to use
        from firedrake.pyplot import triplot

jwallwork23:
Closed by #74.

PR #74, opened 2024-01-27
updated firedrake import for triplot

acse-ej321:
added updated import statement for firdrake.pyplot triplot for existing demos for issue #74

### DIFF
--- a/demos/combining_metrics.py
+++ b/demos/combining_metrics.py
@@ -37,6 +37,7 @@
 import matplotlib.pyplot as plt
 from firedrake import *
 from animate import *
+from firedrake.pyplot import triplot
 
 mesh = UnitSquareMesh(100, 100)
 P1_ten = TensorFunctionSpace(mesh, "CG", 1)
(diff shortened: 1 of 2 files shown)


jwallwork23:
Thanks for this. We'll need to do the same for Goalie, too. The reason the tests fail is that I need to update the Docker container. Will aim to do that soon and then this can go through. https://github.com/pyroteus/animate/pull/47 would be a longer-term solution.

jwallwork23:
By the way, I think the issue you are referring to is #73. When you open the PR, if you start the text block with "Closes #73." then GitHub will link the issue to the PR so that once the PR is merged the issue is automatically closed. Sadly, this only works when you first create the PR, not if you edit the text. However, you can also link them by clicking the cog symbol next to "Development" in the right hand pane.

jwallwork23 (approved):
Okay the good news is that I managed to get the Docker container to be rebuilt in #47 so the tests that were failing now pass. The bad news is that it looks like there's a problem with running in parallel... unrelated to this PR though, so happy to merge.

PR #90, opened 2024-03-13
Allow metric_parameters kwarg to RiemannianMetric

### LINKED ISSUE #89: Metrics parameters argument to RiemannianMetric
How about we pass metric parameters as an argument to `RiemannianMetric` instead of using `set_parameters`? We could then introduce the `update_parameters` method.

ddundo:
Closes #89.

This PR allow us to do `metric = RiemannianMetric(fspace, metric_parameters=mp)` which is equivalent to
```
metric = RiemannianMetric(fspace)
metric.set_parameters(metric_parameters=mp)
```

@jwallwork23 I reordered things in `__init__` because I had to call `self.set_parameters` after `self._plex = plex` but I also think it's nice that now there is no clutter before `super().__init__`, so I think it's clearer and more readable.

### DIFF
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -44,17 +44,11 @@ class RiemannianMetric(ffunc.Function):
             :class:`~.Function`. In this case, the function values are copied. If a
             :class:`~firedrake.mesh.MeshGeometry` is passed here then a tensor
             :math:`\mathbb P1` space is built on top of it.
+        :kwarg metric_parameters: same as for :func:`set_parameters <set_parameters>`.
         """
         if isinstance(function_space, fmesh.MeshGeometry):
             function_space = ffs.TensorFunctionSpace(function_space, "CG", 1)
-        self.metric_parameters = {}
-        self._variable_parameters = {
-            "dm_plex_metric_h_min": firedrake.Constant(1.0e-30),
-            "dm_plex_metric_h_max": firedrake.Constant(1.0e30),
-            "dm_plex_metric_a_max": firedrake.Constant(1.0e5),
-            "dm_plex_metric_boundary_tag": None,
-        }
-        self._variable_parameters_set = False
+        self.metric_parameters = kwargs.pop("metric_parameters", {})
         super().__init__(function_space, *args, **kwargs)
 
         # Check that we have an appropriate tensor P1 function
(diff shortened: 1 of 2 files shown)


animate/metric.py
```diff
@@ -44,17 +44,11 @@ def __init__(self, function_space, *args, **kwargs):
...
             function_space = ffs.TensorFunctionSpace(function_space, "CG", 1)
-        self.metric_parameters = {}
-        self._variable_parameters = {
-            "dm_plex_metric_h_min": firedrake.Constant(1.0e-30),
-            "dm_plex_metric_h_max": firedrake.Constant(1.0e30),
-            "dm_plex_metric_a_max": firedrake.Constant(1.0e5),
-            "dm_plex_metric_boundary_tag": None,
-        }
-        self._variable_parameters_set = False
+        self.metric_parameters = kwargs.pop("metric_parameters", {})
```
jwallwork23, line 51: Please could you make this an explicit kwarg, rather than getting it from `**kwargs`?

animate/metric.py
```diff
@@ -87,6 +81,16 @@ def __init__(self, function_space, *args, **kwargs):
...
 
+        self._variable_parameters = {
+            "dm_plex_metric_h_min": firedrake.Constant(1.0e-30),
+            "dm_plex_metric_h_max": firedrake.Constant(1.0e30),
+            "dm_plex_metric_a_max": firedrake.Constant(1.0e5),
+            "dm_plex_metric_boundary_tag": None,
+        }
+        self._variable_parameters_set = False
+        if self.metric_parameters:
+            self.set_parameters(self.metric_parameters)
```
jwallwork23, line 92: While we're here, I guess it would be worth adding a comment for this code block - something along the lines of "Process spatially variable metric parameters"?

jwallwork23 (changes requested):
Nice, thanks @ddundo. Would you mind either adding a new test or modifying one of the existing ones so that we test passing metric parameters in the new approach? Something in `test/test_metric.py::TestSetParameters`, perhaps?

animate/metric.py
ddundo: I tried that initially:
```
def __init__(self, function_space, metric_parameters={}, *args, **kwargs):
        # ...
        self.metric_parameters = metric_parameters
```
and everything else the same, but this results in 120 tests failing. I can't figure out for the life of me why. So I changed to `kwargs.pop` which I think is identical, and that works. Could you give it a try please?

jwallwork23: Hm yeah I see. Maybe it's something to do with `RiemannianMetric` being a subclass of `Function`. Not a major issue so let's just leave it for now.

animate/metric.py
jwallwork23: Fixed in 3add3e6.

ddundo:
Thanks @jwallwork23, I'll try to figure out how to solve your kwarg suggestion :)

PR #99, opened 2024-03-22
Introduce metric parameters documentation

### LINKED ISSUE #31: Docs of plex parameters
We really need to document what all the `dm_plex_metric_...` parameters mean!

### LINKED ISSUE #98: RiemannianMetric.copy not deep-copying parameters
[L190 in metric.py](https://github.com/pyroteus/animate/blob/bf8e2bb0187cf75c94431765d63d2aba13268dc6/animate/metric.py#L190)  `metric.set_parameters(self.metric_parameters.copy())` doesn't deep copy the `metric_parameters` dictionary. 

MFE:
```
from firedrake import *
from animate.metric import RiemannianMetric


mesh = UnitSquareMesh(3, 3)
P1_ten = TensorFunctionSpace(mesh, "CG", 1)
metric = RiemannianMetric(P1_ten)

x, y = SpatialCoordinate(mesh)
h_min = Function(FunctionSpace(mesh, "CG", 1)).interpolate(x / 2.)
mp = {"h_min": h_min, "h_max": 1.0}
metric.set_parameters(mp)

metric_copy = metric.copy(deepcopy=True)
mp_copy = metric_copy.metric_parameters

print(id(mp) == id(mp_copy))                    # False
print(id(mp["h_min"]) == id(mp_copy["h_min"]))  # True
print(id(mp["h_max"]) == id(mp_copy["h_max"]))  # True
```

jwallwork23:
~~Closes #98.~~

~~Fixes the issue with deep-copying metric parameters by **always** deep-copying these parameters upon setting them. (I can't think of a reason why you wouldn't want this.) Also moves some mechanics behind the scenes and provides a public interface to `metric_parameters`.~~

Edit: See discussion in the original issue.

Closes #31.

At long last!

### DIFF
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -35,6 +35,26 @@ class RiemannianMetric(ffunc.Function):
       https://petsc.org/release/docs/manual/dmplex/#metric-based-mesh-adaptation
     """
 
+    _supported_parameters = (
+        "dm_plex_metric_target_complexity",
+        "dm_plex_metric_h_min",
+        "dm_plex_metric_h_max",
+        "dm_plex_metric_a_max",
+        "dm_plex_metric_p",
+        "dm_plex_metric_gradation_factor",
+        "dm_plex_metric_hausdorff_number",
+        "dm_plex_metric_boundary_tag",
+        "dm_plex_metric_no_insert",
+        "dm_plex_metric_no_swap",
+        "dm_plex_metric_no_move",
+        "dm_plex_metric_no_surf",
+        "dm_plex_metric_num_iterations",
+        "dm_plex_metric_verbosity",
+        "dm_plex_metric_isotropic",
+        "dm_plex_metric_uniform",
+        "dm_plex_metric_restrict_anisotropy_first",
+    )
+
     @PETSc.Log.EventDecorator()
     def __init__(self, function_space, *args, **kwargs):
         r"""
@@ -98,14 +118,29 @@ class RiemannianMetric(ffunc.Function):
         if (el.family(), el.degree()) != ("Lagrange", 1):
             raise ValueError(f"Riemannian metric should be in P1 space, not '{el}'.")
 
-    def _process_parameters(self, metric_parameters):
-        mp = metric_parameters.copy()
+    @staticmethod
+    def _collapse_parameters(metric_parameters):
+        """
+        Account for concise nested dictionary formatting
+        """
+        if "dm_plex_metric" in metric_parameters:
+            for key, value in metric_parameters["dm_plex_metric"].items():
+                metric_parameters["_".join(["dm_plex_metric", key])] = value
+            metric_parameters.pop("dm_plex_metric")
+        return metric_parameters
 
-        # Account for concise nested dictionary formatting
-        if "dm_plex_metric" in mp:
-            for key, value in mp["dm_plex_metric"].items():
-                mp["_".join(["dm_plex_metric", key])] = value
-            mp.pop("dm_plex_metric")
+    def _process_parameters(self, metric_parameters):
+        mp = self._collapse_parameters(metric_parameters.copy())
+
+        # Check all parameters
+        for key in mp:
+            if not key.startswith("dm_plex_metric_"):
+                raise ValueError(
+                    f"Unsupported metric parameter '{key}'."
+                    " Metric parameters must start with the prefix 'dm_plex_metric_'."
+                )
+            if key not in self._supported_parameters:
+                raise ValueError(f"Unsupported metric parameter '{key}'.")
 
         # Spatially varying parameters need to be treated differently
         vp = {}
@@ -192,6 +227,10 @@ class RiemannianMetric(ffunc.Function):
         if self._plex.metricIsIsotropic():
             raise NotImplementedError(
                 "Isotropic metric optimisations are not supported in Firedrake."
+            )
+        if self._plex.metricRestrictAnisotropyFirst():
+            raise NotImplementedError(
+                "Restricting metric anisotropy first is not supported in Firedrake."
             )
 
     @property
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -194,7 +194,7 @@ class RiemannianMetric(ffunc.Function):
             For more detail, see
             https://www.mmgtools.org/mmg-remesher-try-mmg/mmg-remesher-options/mmg-remesher-option-hausd.
         * `boundary_tag`: Mesh boundary tag to restrict attention to during
-            boundary-specific manipulations. Unset by default, which implies all
+            boundary-specific metric manipulations. Unset by default, which implies all
             boundaries are considered. (Note that this parameter does not currently exist
             in the underlying PETSc implementation.)
         * `no_insert`: Boolean flag for turning off node insertion and deletion during
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -133,6 +133,8 @@ class RiemannianMetric(ffunc.Function):
         Set metric parameter values internally. All options have the prefix
         `dm_plex_metric_` and are listed below (with prefix dropped for brevity):
 
+        * `target_complexity`: Strictly positive target metric complexity value. No
+            default - **must be set**.
         * `h_min`: Minimum tolerated metric magnitude, which allows approximate control
             of minimum element size in the adapted mesh. Supports :class:`~.Constant`
             and :class:`~.Function` input, as well as :class:`float`. Default: 1.0e-30.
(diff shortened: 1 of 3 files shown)


animate/metric.py
```diff
@@ -128,19 +133,58 @@ def _process_parameters(self, metric_parameters):
...
+        * `p`: :math:`L^p` normalisation order. Supports ``np.inf`` as well as
+            :class:`float` values from :math:`[0,\infty)`. Default: 1.0.
+        * `target_complexity`: Strictly positive target metric complexity value. No
+            default - **must be set**.
+        * `gradation_factor`: Maximum ratio by which adjacent edges in the adapted mesh
+            may differ. Default: 1.3.
+        * `hausdorff_number`: Spatial scale factor for the problem. Default: 0.01.
+        * `boundary_tag`: Mesh boundary tag to restrict attention to during
+            boundary-specific manipulations. Unset by default, which implies all
+            boundaries are considered. (Note that this parameter does not currently exist
```
acse-ej321, line 159: It might be good to add how this parameter currently interacts with a similar flag which can be set in a mesh originating from a `.gmsh`, as Firedrake does seem to preserve that flag when communicating between PETSc and MMG. Is this a different flag altogether?

animate/metric.py
```diff
@@ -128,19 +133,58 @@ def _process_parameters(self, metric_parameters):
...
+        * `no_insert`: Boolean flag for turning off node insertion and deletion during
+            adaptation. Default: False.
+        * `no_swap`: Boolean flag for turning off edge and face swapping during
+            adaptation. Default: False.
+        * `no_move`: Boolean flag for turning off node movement during adaptation.
+            Default: False.
+        * `no_surf`: Boolean flag for turning off surface modification during adaptation.
+            Default: False.
+        * `num_iterations`: Number of adaptation iterations in the parallel case.
+            Default: 3.
```
acse-ej321, line 170: The description leaves open the question of what happens in the non-parallel case. Is there only one adaptation iteration to the metric in the non-parallel case?

animate/metric.py
```diff
@@ -181,13 +232,14 @@ def copy(self, deepcopy=False):
...
-            :class:`~.RiemannianMetric` will share the dof values.
+        :kwarg deepcopy: If ``True``, the new metric will allocate new space and copy
+            values. If ``False`` (default) then the new metric will share the DoF values.
+        :type deepcopy: :class:`bool`
         :return: a copy of the metric with the same parameters set
+        :rtype: :class:`~.RiemannianMetric`
         """
         metric = type(self)(super().copy(deepcopy=deepcopy))
-        metric.set_parameters(self.metric_parameters.copy())
+        metric.set_parameters(self.metric_parameters)
```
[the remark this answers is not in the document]
ddundo: I see that you put in a unit test for this which passes for me. But when I do the following:
```
from firedrake import *
from animate.metric import RiemannianMetric


mesh = UnitSquareMesh(1, 1)
metric1 = RiemannianMetric(TensorFunctionSpace(mesh, "CG", 1))

P1 = FunctionSpace(mesh, "CG", 1)
mp1 = {"h_min": Function(P1).assign(1.0), "h_max": 1.0}
metric1.set_parameters(mp1)

metric2 = metric1.copy(deepcopy=True)
mp2 = metric2.metric_parameters

print(id(mp1) == id(mp2))                    # False
print(id(mp1["h_min"]) == id(mp2["h_min"]))  # True
print(id(mp1["h_max"]) == id(mp2["h_max"]))  # True

mp1["h_min"].assign(999)
mp1["h_max"] = 999

print(mp1["h_min"].dat.data)   # 999
print(mp1["h_max"])            # 999
print("-------------------")
print(mp2["h_min"].dat.data)  # 999  - WRONG
print(mp2["h_max"])           # 1.0

print(id(mp1) == id(mp2))                    # False
print(id(mp1["h_min"]) == id(mp2["h_min"]))  # True
print(id(mp1["h_max"]) == id(mp2["h_max"]))  # False
```

it doesn't seem to deep-copy variable parameters. I thought it might make a difference if we define the `Function` outside of the dictionary, i.e. sth like
```
h_min = Function(...)
mp = {"h_min": h_min}
```
but that doesn't seem to make a difference. So I'm not sure why the test passes and this doesn't... I don't see a difference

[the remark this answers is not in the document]
ddundo: Ah I see... It's because in this example I lazily wrote `"h_min"` instead of "`dm_plex_metric_h_min"` (which I always do).

Now I feel like having two accepted ways of naming parameters is just increasing chances for issues popping up. Why not just stick with one to make sure the code is bulletproof?

animate/metric.py
```diff
@@ -128,19 +133,58 @@ def _process_parameters(self, metric_parameters):
         return mp, vp
 
     def set_parameters(self, metric_parameters={}):
```
ddundo, line 135: I think it would be useful to raise a warning if the user tries to set a parameter which doesn't exist or isn't supported. In case of typos - which I think are quite likely since parameter names are quite long, like `dm_plex_metric_target_complexity`...

ddundo: It happened to me multiple times actually and it's always annoying to debug

ddundo (changes requested):
Thanks for this @jwallwork23! Minor comments from me :)

animate/metric.py
jwallwork23: See 8834c49.

animate/metric.py
```diff
@@ -128,19 +133,58 @@ def _process_parameters(self, metric_parameters):
...
+            of maximum element anisotropy in the adapted mesh. Supports
+            :class:`~.Constant` and :class:`~.Function` input, as well as :class:`float`.
+            Default: 1.0e+05.
+        * `p`: :math:`L^p` normalisation order. Supports ``np.inf`` as well as
+            :class:`float` values from :math:`[0,\infty)`. Default: 1.0.
+        * `target_complexity`: Strictly positive target metric complexity value. No
+            default - **must be set**.
+        * `gradation_factor`: Maximum ratio by which adjacent edges in the adapted mesh
+            may differ. Default: 1.3.
+        * `hausdorff_number`: Spatial scale factor for the problem. Default: 0.01.
```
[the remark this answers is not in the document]
jwallwork23: Addressed in 8da91ac.

animate/metric.py
```diff
@@ -128,19 +133,58 @@ def _process_parameters(self, metric_parameters):
...
+            of maximum element size in the adapted mesh. Supports :class:`~.Constant`
+            and :class:`~.Function` input, as well as :class:`float`. Default: 1.0e+30.
+        * `a_max`: Maximum tolerated metric anisotropy, which allows approximate control
+            of maximum element anisotropy in the adapted mesh. Supports
+            :class:`~.Constant` and :class:`~.Function` input, as well as :class:`float`.
+            Default: 1.0e+05.
+        * `p`: :math:`L^p` normalisation order. Supports ``np.inf`` as well as
+            :class:`float` values from :math:`[0,\infty)`. Default: 1.0.
+        * `target_complexity`: Strictly positive target metric complexity value. No
+            default - **must be set**.
```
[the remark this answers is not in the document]
jwallwork23: Addressed in c8469a3fcb58ac669c2f6d8bba42a671438c2da4.

animate/metric.py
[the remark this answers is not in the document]
jwallwork23: I'm ignoring the first comment based on our discussion on the original issue. See 3ae4d9e900362bd0adb22db1848c1623c6b769d6 for my addressing your latest comment.

animate/metric.py
jwallwork23: Good idea. Fixed in 3ae4d9e900362bd0adb22db1848c1623c6b769d6.

animate/metric.py
```diff
@@ -150,6 +235,17 @@ def set_parameters(self, metric_parameters={}):
             raise NotImplementedError(
                 "Isotropic metric optimisations are not supported in Firedrake."
             )
+        if self._plex.metricRestrictAnisotropyFirst():
+            raise NotImplementedError(
+                "Restricting metric anisotropy first is not supported in Firedrake."
+            )
```
jwallwork23, line 241: This should be easy enough to implement (just tweak the arguments to the `enforce_spd` call in the definition of `normalise`. However, I'd prefer to address it in a separate issue rather than here, considering the impact with some tests.

acse-ej321 (approved):

animate/metric.py
jwallwork23: See #100.

jwallwork23:
You happy with the changes @ddundo?

ddundo (approved):
Thanks @jwallwork23 :)

PR #105, opened 2024-04-14
Port interpolation functionality over from Goalie

jwallwork23:
Closes #94.

Also introduces a hack to temporarily avoid #92.

PR #106, opened 2024-04-15
Drop duplicated tests

### LINKED ISSUE #92: Parallel Hessian tests fail locally
I have two tests failing locally, even on `main`:
```
test/test_parallel.py::test_hessian_bowl_np2[2]
test/test_parallel.py::test_hessian_bowl_np2[3]
```
The error messages are assertion errors for the metric against its expected value.

jwallwork23:
Closes #92.

It seems I accidentally added the `TestTransfer` unit tests twice in #105 - this PR drops the copies.

I also found that #92 is fixed locally following a fresh Firedrake install so the temporary hack is no longer needed :)


ddundo (approved):
Sorry for missing the duplicate test!

jwallwork23:
> Sorry for missing the duplicate test!

No worries

Issue #114, opened 2024-04-27
Adjoint transfer tests are incorrect


jwallwork23:
If different meshes are used instead of the trivial ones currently used in `TestTransfer` then we get test failures for the adjoint transfer operators. The tests assert that applying the adjoint transfer to a source should give the same result as a forward transfer, but this is nonsense.

jwallwork23:
Two of the tests are actually fine (adjoint transfers between matching spaces) and are re-enabled in #187.

However, the tests for different spaces on the same mesh aren't and need more work. We should write down what the transfer matrices look like for very small, simple examples and use those to work out what the adjoints should be.

Issue #113, opened 2024-04-27
Adjoint interpolation operator is incorrect


jwallwork23:
The adjoint version of the `interpolate` operator simply applies the standard `interpolate`. This is incorrect and needs to be implemented by transposing the interpolation operator.

PR #115, opened 2024-04-27
Enable mass lumping for projection operator

### LINKED ISSUE #95: Allow mass lumping in projection
Mass lumping the mesh-to-mesh projection operator will allow us to bound the interpolator for P1 fields.

jwallwork23:
Closes #95.

This PR allows the projection operator (and its adjoint) to be mass lumped. For P1 spaces, this implies that extrema are preserved following the transfer.

While developing these changes, I noticed #113 and #114, so the associated functionality and tests are turned off here.


jwallwork23:
Still work to do to address the issues referenced above, but hopefully this is sufficient for your purposes @ddundo.

ddundo:
Thanks very very much @jwallwork23! And sorry for the issues you raised - I tried to warn you about my incompetence in https://github.com/mesh-adaptation/goalie/pull/128#issuecomment-2001392132 :)

This actually makes quite a significant difference in my case, which I didn't expect. I will bring it up on the next meeting :)

jwallwork23:
TODO: Make sure project is the default, not interpolate.

jwallwork23:
Hopefully this is useful @ddundo! New `minimal_diffusion` option. A few tweaks still required, including plotting. However, I thought you might find the code of use.

ping_pong-quantities_mindiff

ping_pong-final

ddundo:
Thanks a lot for doing this @jwallwork23! Looks great from what you show here!

I couldn't use it directly in my case because I have a vector field, so I tried to modify it by considering each component as a separate scalar field, but then didn't get very good results. So I assume something more sophisticated is necessary :)

As a side note... I am also confused how tests pass for you here since you didn't merge the commit for #117 here... I had to do it locally.

jwallwork23:
> Thanks a lot for doing this @jwallwork23! Looks great from what you show here!

No worries - I had a long train journey and this was an interesting puzzle to solve during it.

> I couldn't use it directly in my case because I have a vector field, so I tried to modify it by considering each component as a separate scalar field, but then didn't get very good results. So I assume something more sophisticated is necessary :)

Ah yes, the vector extension will have to come in a later PR I think. Shouldn't be too hard, though.

> As a side note... I am also confused how tests pass for you here since you didn't merge the commit for #117 here... I had to do it locally.

Hm yeah good point. Pass.

jwallwork23:
I'll probably rename the `minimal_diffusion` kwarg as `bounded` (as in the paper) and drop the `lumped` option. (I don't think there would be a case where you'd want to use lumping on its own without the post-processing.)

jwallwork23:
@ddundo This is now ready for review. Minimal diffusion correction taken out for now and marked as future work in #124. Note also #123.

Do you know how to get the colourbars formatted correctly?

ddundo:
Thanks a lot for this @jwallwork23! The problem with the colourbars is that the levels in `tricontourf` and `ticks` in colorbar are different. So if you do
```
fig, axes = plt.subplots(ncols=2, nrows=2, figsize=(10, 10))
levels = [-0.05] + [0.15 * i for i in range(1, 8)]
for ax, f in zip(axes.flatten(), (sensor, f_interp, f_proj, f_bounded)):
    im = tricontourf(f, axes=ax, levels=levels)
    ax.axis(False)
    ax.set_aspect(1)
    fig.colorbar(im, ax=ax, ticks=levels)
# fig.colorbar(im, ax=axes, ticks=levels)  # or replace above with this for a single colourbar
plt.savefig("ping_pong-final.jpg")
```
you get this
image

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
+# 'Ping pong' interpolation experiment
+# ====================================
+#
+# In this demo, we perform a 'ping poing test', which amounts to interpolating a given
```
ddundo, line 4:
```suggestion
# In this demo, we perform a 'ping pong test', which amounts to interpolating a given
```

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
...
+
+import matplotlib.pyplot as plt
+from firedrake import *
+from firedrake.pyplot import tricontourf, triplot
+
+from animate import *
+
+# Consider two different meshes of the unit square: mesh A, which is laid out
+# :math:`20\times25` with diagonals from top left to bottom right and mesh B
+# :math:`20\times20` with diagonals from bottom left to top right. ::
```
ddundo, line 19: I initially missed that the meshes have different resolutions. Maybe add "a finer mesh A ... and a coarser mesh B"?

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
...
+
+mesh_A = UnitSquareMesh(20, 25, diagonal="left")
+mesh_B = UnitSquareMesh(20, 20, diagonal="right")
+
+fig, axes = plt.subplots(ncols=2, figsize=(10, 5))
+for i, mesh in enumerate((mesh_A, mesh_B)):
+    triplot(mesh, axes=axes[i])
+    axes[i].set_title(f"Mesh {'B' if i else 'A'}")
```
ddundo, line 27: `axes[i].set_title(f"Mesh {'AB'[i]}")` is maybe a bit less confusing :)

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
...
+
+fig, axes = plt.subplots()
+tricontourf(sensor, axes=axes)
+axes.set_title("Source function")
+axes.axis(False)
+axes.set_aspect(1)
+plt.savefig("ping_pong-source_function.jpg")
+
+# To start with, let's consider the `interpolate` approach, which point evaluates the
+# input field at the locations where degrees of freedom of the target function space.
```
ddundo, line 53: A verb is missing here. "... function space ARE"? Or "at the locations OF degrees of freedom..."?

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
...
+mesh_A = UnitSquareMesh(20, 25, diagonal="left")
+mesh_B = UnitSquareMesh(20, 20, diagonal="right")
+
+fig, axes = plt.subplots(ncols=2, figsize=(10, 5))
+for i, mesh in enumerate((mesh_A, mesh_B)):
+    triplot(mesh, axes=axes[i])
+    axes[i].set_title(f"Mesh {'B' if i else 'A'}")
+    axes[i].axis(False)
+    axes[i].set_aspect(1)
+plt.savefig("ping_pong-meshes.jpg")
```
ddundo, line 30: I'd add `bbox_inches="tight"` to all figures here (and all current/future demos too btw).

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
...
+
+# .. figure:: ping_pong-quantities_interpolate.jpg
+#    :figwidth: 90%
+#    :align: center
+#
+# The first plot shows the integral of the field (the 'mass') as the interpolation
+# iterations progress. Note that the mass drops steeply, before leveling out after
+# around 50 iterations. Minimum values are perfectly maintained for this example,
+# staying at zero for all iterations. Maximum values, however, decrease for several
+# iterations before leveling out after around 30 iterations.
```
ddundo, line 89: I think it's useful to say here what you said on the meeting too: that this is due to the minimum values being achieved at the boundary, while the maximum is in the middle (and then 20x20 <-> 20x25 resolution shifting that)

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
...
+
+# .. figure:: ping_pong-quantities_project.jpg
+#    :figwidth: 90%
+#    :align: center
+#
+# The first subfigure shows that mass is indeed conserved by the `project` approach,
+# unlike the `interpolate` approach. However, contrary to the `interpolate` approach
+# not introducing new extrema, the `project` approach can be seen to introduce new
+# minimum values. No new maximum values are introduced, although the maximum values do
+# decrease slightly.
```
ddundo, line 149: As for interpolate, I think it would be great to explicitly say why this is

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
...
+ax.set_title("Project")
+ax = axes[1][1]
+colourbars.append(fig.colorbar(tricontourf(f_bounded, axes=ax), ax=ax))
+ax.set_title("Bounded")
+for i in range(2):
+    for j in range(2):
+        axes[i][j].axis(False)
+        axes[i][j].set_aspect(1)
+for cbar in colourbars:
+    cbar.set_ticks([-0.05, 0.15, 0.3, 0.45, 0.6, 0.75, 0.9, 1.05])
```
ddundo, line 210: I left a separate comment on how to deal with colourbars.

I also think it would be nice to add a contour at 0., so that we can actually see in the `project` figure that it introduces a new minimum.

demos/ping_pong.py
```diff
@@ -0,0 +1,233 @@
...
+plt.savefig("ping_pong-final.jpg")
+
+# .. figure:: ping_pong-final.jpg
+#    :figwidth: 90%
+#    :align: center
+#
+# Whilst the first two approaches clearly resemble the input field, the bounded version
+# gives a much poorer representation. As such, we find that, whilst the bounded
```
ddundo, line 218: Again I'd add a quick comment that this is due to added numerical diffusion

ddundo (approved):
Thanks very much again for this! I just left minor comments :)

demos/ping_pong.py
ddundo: And I'd like to remove the 1.05 tick to remove any suggestion that there might be any values >1.

jwallwork23:
Thanks for the thorough review @ddundo. I think I addressed all points in b775043655245bb4c1bd1782a9fb2d28099ca945.

ddundo (approved):
Looks good! Sorry for nitpicking :)

Oh and I also ran the tests both on my local install and on the docker container, and tests pass for me. I also updated firedrake.

jwallwork23:
Thanks again @ddundo. Pushed a minor change that fixes the CI, not controversial.

PR #116, opened 2024-05-04
Fixes for adapting in serial PR

ddundo:
Here are some changes that I think address all issues raised in the original PR #72:
- I am creating the checkpoint dir using `tempfile.mkdtemp`, as suggested in [this comment](https://github.com/mesh-adaptation/animate/pull/72#discussion_r1523844000)
- Instead of the checkpoint file name, `save_checkpoint` and `load_checkpoint` now take the checkpoint file path as an argument. The main reason for this is to tidy up the code since I don't really see a benefit of doing it the other way around. I also think it was confusing. But I'm happy to revert this!
- I moved `get_checkpoint_dir` from `utility.py` to `checkpointing.py`. I'm not sure why we'd have checkpointing-related things outside of `checkpointing.py`? :)
- Added a test for attempting to load a checkpoint file that doesn't exist

### DIFF
--- a/animate/adapt.py
+++ b/animate/adapt.py
@@ -1,6 +1,6 @@
 import abc
 import os
-import time
+from shutil import rmtree
 
 import firedrake.checkpointing as fchk
 import firedrake.functionspace as ffs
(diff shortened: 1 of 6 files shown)


ddundo:
I opened #117 for the error raised in the failing tests. Should be ready for review after that is addressed :)

jwallwork23 (approved):
This is great, thanks @ddundo!

PR #119, opened 2024-05-09
#81 add import interpolate from __future__ and update function calls

### LINKED ISSUE #81: Avoid interpolation deprecation warning
Do the same in animate as in https://github.com/pyroteus/goalie/issues/105

acse-ej321:
Closes #81

Updates calls to `interpolate` in animate to be consistent with `__future__` implementation in Firedrake.


jwallwork23 (approved):
LGTM

Issue #108, opened 2024-04-17
Fix B and C linting rules


ddundo:
The current linter does not seem to catch "B" and "C" rules even though they are enabled in `.flake8`. Running the ruff linter caught the following:

```
animate/interpolation.py:218:5: C901 `clement_interpolant` is too complex (19 > 10)
animate/math.py:26:9: B007 Loop control variable `i` not used within loop body
animate/metric.py:168:48: B006 Do not use mutable data structures for argument defaults
animate/quality.py:88:18: C408 Unnecessary `dict` call (rewrite as a literal)
animate/utility.py:96:5: C901 `norm` is too complex (13 > 10)
animate/utility.py:152:17: B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
animate/utility.py:168:5: C901 `errornorm` is too complex (12 > 10)
test/test_adapt.py:35:13: B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
test/test_setup.py:33:53: B006 Do not use mutable data structures for argument defaults
Found 9 errors.
```

PR #120, opened 2024-05-09
Fix B linting rules

ddundo:
Addresses the B rules in #108. Doesn't introduce any functional changes.

I can't connect to my server today so I thought I'd take care of some easy issues.


jwallwork23 (approved):
Seems reasonable, thanks!

PR #125, opened 2024-05-15
Fixes for ping pong demo

jwallwork23:
This PR includes various formatting fixes for the ping pong demo, e.g., math rendering, citation.

It also ensures that all of Animate's modules define `__all__`, otherwise the things they import will get bundled along with `from animate import *`. The issue here was that `quality` was importing `interpolate` from `firedrake.__future__` and this was replacing the `interpolate` function defined by Animate.

### DIFF
--- a/animate/interpolation.py
+++ b/animate/interpolation.py
@@ -29,6 +29,9 @@ def transfer(source, target_space, transfer_method="project", **kwargs):
     function spaces defined on different meshes and for the adjoint interpolation
     operator when applied to :class:`firedrake.cofunction.Cofunction`\s.
 
+    Note that the "interpolate" option works straightforwardly with MPI parallelism,
+    whereas the "project" option can be difficult to set up to make use of this.
+
     :arg source: the function to be transferred
     :type source: :class:`firedrake.function.Function` or
         :class:`firedrake.cofunction.Cofunction`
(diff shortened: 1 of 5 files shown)


jwallwork23:
Note https://github.com/mesh-adaptation/mesh-adaptation-docs/issues/27.

ddundo (approved):
Sorry for missing these - I indeed wasn't rebuilding docs every time. I will from now on.

This looks good to me! Just 2 minor comments:
1. The y-label of the middle subplot ("minimum") overlaps with the left subplot in two figures.
2. It occurred to me that it would probably be worth mentioning that `interpolate` can be used in parallel, while `project` cannot

jwallwork23:
Thanks @ddundo!

Website updated: https://mesh-adaptation.github.io/demos/ping_pong.py.html.

Will merge provided tests pass.

ddundo:
@jwallwork23 sorry, I just realised that you aren't showing the first figure (the one with the two meshes). Is that intentional?

jwallwork23:
> @jwallwork23 sorry, I just realised that you aren't showing the first two figures (the two meshes and source function). Is that intentional?

Ah okay, no that wasn't intentional. Let's open it as an issue and address this along with a future issue, rather than opening a PR especially.

PR #134, opened 2024-05-22
Modify and run demos directly during testing

### LINKED ISSUE #110: Run demos directly during testing
Some of the code coverage information is missing because we are running the demos using subprocesses.

Perhaps we could instead run them directly with Pytest and have the Pytest configuration delete any outputs.

### LINKED ISSUE #111: Use `sed` instead of `ANIMATE_REGRESSION_TEST`
Rather than confusing the demo content with environment variables, we should apply specific modifications with `sed` when running the test suite.

ddundo:
Closes #110. Closes #111.

See https://github.com/mesh-adaptation/goalie/pull/199

### DIFF
--- a/demos/ping_pong.py
+++ b/demos/ping_pong.py
@@ -6,7 +6,6 @@
 # method. The purpose of this experiment is to assess the properties of different
 # interpolation methods. ::
 
-import os
 
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


jwallwork23 (approved):
Thanks @ddundo!

PR #138, opened 2024-06-25
Enable C linting rules

### LINKED ISSUE #108: Fix B and C linting rules

ddundo:
Closes #108.

After #120, the remaining linting errors are:
```
animate/interpolation.py:296:5: C901 `clement_interpolant` is too complex (19 > 10)
animate/quality.py:91:18: C408 Unnecessary `dict` call (rewrite as a literal)
animate/utility.py:97:5: C901 `norm` is too complex (13 > 10)
animate/utility.py:169:5: C901 `errornorm` is too complex (12 > 10)
```

This PR addresses all of them apart from the final one, to which I added `noqa: C901`. The complexity mostly comes from the `if isinstance` checks so I don't think it's that bad to begin with. And I guess I could replace those with `assert` but I would rather not do that. @jwallwork23 do you see a nice way to simplify it please? :)

### DIFF
--- a/animate/interpolation.py
+++ b/animate/interpolation.py
@@ -326,15 +326,15 @@ def clement_interpolant(source, target_space=None, boundary=False):
     # Process target space
     Vt = target_space
     if Vt is None:
-        if rank == 0:
-            Vt = firedrake.FunctionSpace(mesh, "CG", 1)
-        elif rank == 1:
-            Vt = firedrake.VectorFunctionSpace(mesh, "CG", 1)
-        else:
-            Vt = firedrake.TensorFunctionSpace(mesh, "CG", 1)
+        Vt_ranks = {
+            0: firedrake.FunctionSpace(mesh, "CG", 1),
+            1: firedrake.VectorFunctionSpace(mesh, "CG", 1),
+            2: firedrake.TensorFunctionSpace(mesh, "CG", 1),
+        }
+        Vt = Vt_ranks[rank]
     elif isinstance(Vt, FiredrakeDualSpace):
         Vt = Vt.dual()
-    elif not isinstance(Vt, FiredrakeDualSpace):
+    else:
         is_cofunction = False
     Vt_e = Vt.ufl_element()
     if not (Vt_e.family() == "Lagrange" and Vt_e.degree() == 1):
@@ -346,14 +346,12 @@ def clement_interpolant(source, target_space=None, boundary=False):
     P1 = firedrake.FunctionSpace(mesh, "CG", 1)
 
     # Determine target domain
-    if rank == 0:
-        tdomain = "{[i]: 0 <= i < t.dofs}"
-    elif rank == 1:
-        tdomain = f"{{[i, j]: 0 <= i < t.dofs and 0 <= j < {dim}}}"
-    else:
-        tdomain = (
-            f"{{[i, j, k]: 0 <= i < t.dofs and 0 <= j < {dim} and 0 <= k < {dim}}}"
-        )
+    tdomain_ranks = {
+        0: "{[i]: 0 <= i < t.dofs}",
+        1: f"{{[i, j]: 0 <= i < t.dofs and 0 <= j < {dim}}}",
+        2: f"{{[i, j, k]: 0 <= i < t.dofs and 0 <= j < {dim} and 0 <= k < {dim}}}",
+    }
+    tdomain = tdomain_ranks[rank]
 
     # Compute the patch volume at each vertex
     if not boundary:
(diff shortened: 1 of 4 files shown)


animate/interpolation.py
```diff
@@ -326,15 +326,15 @@ def clement_interpolant(source, target_space=None, boundary=False):
...
-        elif rank == 1:
-            Vt = firedrake.VectorFunctionSpace(mesh, "CG", 1)
-        else:
-            Vt = firedrake.TensorFunctionSpace(mesh, "CG", 1)
+        Vt_ranks = {
+            0: firedrake.FunctionSpace(mesh, "CG", 1),
+            1: firedrake.VectorFunctionSpace(mesh, "CG", 1),
+            2: firedrake.TensorFunctionSpace(mesh, "CG", 1),
+        }
+        Vt = Vt_ranks[rank]
```
jwallwork23, line 334: This is less efficient because now we construct 3 spaces rather than 1. How about this?
```suggestion
    Vt = {
        0: firedrake.FunctionSpace,
        1: firedrake.VectorFunctionSpace,
        2: firedrake.TensorFunctionSpace,
    }[rank](mesh, "CG", 1)
    ```

File: animate/interpolation.py [hunk 2]
```diff
@@ -346,14 +346,12 @@ def clement_interpolant(source, target_space=None, boundary=False):
...
-    else:
-        tdomain = (
-            f"{{[i, j, k]: 0 <= i < t.dofs and 0 <= j < {dim} and 0 <= k < {dim}}}"
-        )
+    tdomain_ranks = {
+        0: "{[i]: 0 <= i < t.dofs}",
+        1: f"{{[i, j]: 0 <= i < t.dofs and 0 <= j < {dim}}}",
+        2: f"{{[i, j, k]: 0 <= i < t.dofs and 0 <= j < {dim} and 0 <= k < {dim}}}",
+    }
+    tdomain = tdomain_ranks[rank]
```
jwallwork23, line 354: Do we need to keep this dictionary in memory? Would the linter not be okay with this?
```suggestion
    tdomain = {
        0: "{[i]: 0 <= i < t.dofs}",
        1: f"{{[i, j]: 0 <= i < t.dofs and 0 <= j < {dim}}}",
        2: f"{{[i, j, k]: 0 <= i < t.dofs and 0 <= j < {dim} and 0 <= k < {dim}}}",
    }[rank]
```

jwallwork23: Similarly below.

animate/utility.py
```diff
@@ -154,19 +153,20 @@ def norm(v, norm_type="L2", condition=None, boundary=False):
...
-            integrand = ufl.inner(v, v) + ufl.div(v) * ufl.div(v)
-        elif norm_type.lower() == "hcurl":
-            integrand = ufl.inner(v, v) + ufl.inner(ufl.curl(v), ufl.curl(v))
+        elif norm_type.lower() in ("h1", "hdiv", "hcurl"):
+            integrands_dict = {
+                "h1": ufl.inner(v, v) + ufl.inner(ufl.grad(v), ufl.grad(v)),
+                "hdiv": ufl.inner(v, v) + ufl.div(v) * ufl.div(v),
+                "hcurl": ufl.inner(v, v) + ufl.inner(ufl.curl(v), ufl.curl(v)),
+            }
+            integrand = integrands_dict[norm_type.lower()]
```
jwallwork23, line 162: Again here. Also, these could be functions rather than passing `v` to them all.

jwallwork23 (changes requested):
Thanks for this @ddundo. This linter option seems a bit over-the-top to me, to be honest. But I have a few suggestions that could improve things a bit.

ddundo:
> Thanks for this @ddundo. This linter option seems a bit over-the-top to me, to be honest. But I have a few suggestions that could improve things a bit.

Thanks @jwallwork23 - very good points! And before I address them, would you prefer if I added C901 to the ignore list if you think it's over-the-top? Then I could revert back to how it was if you'd like

jwallwork23:
> Thanks @jwallwork23 - very good points! And before I address them, would you prefer if I added C901 to the ignore list if you think it's over-the-top? Then I could revert back to how it was if you'd like

No it's okay, let's stick with it. It's better to go with linter recommendations rather than ignoring them.

animate/interpolation.py
ddundo: Thanks! Done as you suggested in 2946fc8a3241de5d1fa4732cda88d6a1ad16c0cf

animate/interpolation.py
ddundo: Both done in 2946fc8a3241de5d1fa4732cda88d6a1ad16c0cf

animate/utility.py
ddundo: Done as you suggested in 2946fc8a3241de5d1fa4732cda88d6a1ad16c0cf

ddundo:
Thanks @jwallwork23 - this tidies it up nicely :) ready for review again

jwallwork23 (approved):

Issue #140, opened 2024-07-07
Mysterious CI crash


jwallwork23:
The automated CI testing failed with a mysterious "exit 1" error last night:
https://github.com/mesh-adaptation/animate/actions/runs/9823758867/job/27121988139

jwallwork23:
Duplicate of #136.

PR #142, opened 2024-07-07
Truncate infinite degree when passing to PETSc

### LINKED ISSUE #141: Fix handling of infinite degree
Goalie's automated CI hit a PETSc error handling infinite normalisation degree last night:
https://github.com/mesh-adaptation/goalie/actions/runs/9823768044/job/27122011016

jwallwork23:
Closes #141.

It seems PETSc no longer likes being passed `np.inf`. Note that we aren't actually using PETSc's metric normalisation funcationality anyway so this isn't a big deal.

### DIFF
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -229,9 +229,8 @@ class RiemannianMetric(ffunc.Function):
         self._variable_parameters.update(vp)
 
         mp = self._metric_parameters.copy()
-        if "dm_plex_metric_p" in mp:
-            if mp["dm_plex_metric_p"] == np.inf:
-                mp["dm_plex_metric_p"] = 1.79769e308
+        if mp.get("dm_plex_metric_p") == np.inf:
+            mp["dm_plex_metric_p"] = 1.79769e308
 
         # Pass parameters to PETSc
         with OptionsManager(mp, "").inserted_options():
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -228,8 +228,13 @@ class RiemannianMetric(ffunc.Function):
         self._metric_parameters.update(mp)
         self._variable_parameters.update(vp)
 
+        mp = self._metric_parameters.copy()
+        if "dm_plex_metric_p" in mp:
+            if mp["dm_plex_metric_p"] == np.inf:
+                mp["dm_plex_metric_p"] = 1.79769e308
+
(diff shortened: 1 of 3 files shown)


jwallwork23:
CI hits #140; will need to do some GitHub Actions debugging with [tmate](https://tmate.io).

jwallwork23:
Can confirm the Goalie tests do pass locally now, though.

jwallwork23:
Hm these tests all passed locally for me. Will have to do an interactive tmate debugging session.

jwallwork23:
Okay I finally got round to debugging this with tmate. There are C++ compilation issues because PyOP2 can't find Eigen. Whilst Firedrake's Docker container apparently no longer installs Eigen, it should be installed when we reconfigure PETSc. There may be an issue with the fact the other packages are stored in `${PETSC_DIR}/packages` rather than `${PETSC_DIR}/${PETSC_ARCH}/externalpackages`.

jwallwork23:
Reproduced the error running the Docker image locally, too:
```
/home/firedrake/firedrake/.cache/pyop2/5a/b01b4c6a652cacd59181a068f880f3_p49.cpp:4:10: fatal error: Eigen/Dense: No such file or directory
    4 | #include <Eigen/Dense>
      |          ^~~~~~~~~~~~~
compilation terminated.
```

jwallwork23:
Now works thanks to https://github.com/mesh-adaptation/mesh-adaptation-docs/pull/40.

animate/metric.py
```diff
@@ -228,8 +228,13 @@ def set_parameters(self, metric_parameters=None):
         self._metric_parameters.update(mp)
         self._variable_parameters.update(vp)
 
+        mp = self._metric_parameters.copy()
+        if "dm_plex_metric_p" in mp:
+            if mp["dm_plex_metric_p"] == np.inf:
```
ddundo, line 233: Could simplify and do this in one line `if "dm_plex_metric_p" in mp and mp["dm_plex_metric_p"] == np.inf:`

ddundo (approved):
Thanks! Just a nitpicky comment :)

animate/metric.py
jwallwork23: Addressed in 5f99eb0.

PR #147, opened 2024-09-15
Expand workflows: citation check and changed-files

ddundo:
See https://github.com/mesh-adaptation/goalie/pull/197


jwallwork23 (approved):
Thanks @ddundo!

PR #148, opened 2024-09-15
Reusable workflows

ddundo:
See https://github.com/mesh-adaptation/mesh-adaptation-docs/issues/45.


ddundo:
I also checked that the citation check is triggered, see https://github.com/mesh-adaptation/animate/actions/runs/10872741835/job/30168109254 (I reset and force-pushed that commit)

jwallwork23 (approved):
This is great, thanks @ddundo.

PR #154, opened 2024-10-15
Bubble shear demo improvements

### LINKED ISSUE #153: Improvements to bubble shear demo
image

- [x] Vast majority of effort in the bubble shear demo goes into interpolating velocity `u.interpolate(velocity_expression(mesh, t))`
  - The reason is that `t` in `velocity_expression` is hardcoded as a `float` so a lot of time is spent recompiling it. Instead, it should be wrapped in `Constant` (or `Function` of R). This made it much faster. 
  - Doing `u.dat.data[:] = u_numpy`, where `u_numpy` is the velocity computed with `numpy` also works and is actually faster than the above, but it's not very firedrake-y, so I went with the above method
- [x] The first computed metric in metric advection should be normalised I think
- [x] Wrong description in Line 416
- [x] Mention explicit timestepping and CFL condition in the exercise rubric
- [ ] Recompute initial condition on the first adapted mesh, rather than projecting the one computed on the uniform mesh. This may actually give a slightly higher error in the end - put it in the exercise rubric?

ddundo:
Closes #153.

### DIFF
--- a/demos/bubble_shear.py
+++ b/demos/bubble_shear.py
@@ -501,6 +501,8 @@ print(
 # subintervals do the two algorithms produce most similar and most different meshes?
 # Experiment with different metric parameters, different adaptation frequencies, and
 # even different velocity fields to further explore the capabilities and limitations of
-# the algorithms presented above.
+# the algorithms presented above. Another interesting experiment would be to compare
+# the impact of switching to an explicit time integration and using a smaller timestep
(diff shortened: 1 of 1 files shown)


ddundo:
Testing the demo still takes quite a while, but is much faster: 49.33s compared to 83.51s on the main branch

jwallwork23 (approved):
All looks good to me, thanks.

PR #169, opened 2024-12-21
Drop lint ignores except in demos

### LINKED ISSUE #168: Drop lint ignores except in demos
We are currently ignoring a few lint checks for no good reason. These should all be enabled.

See https://github.com/mesh-adaptation/movement/pull/126 for a way to ignore lint checks only in the demos.

jwallwork23:
Closes #168.

### DIFF
--- a/animate/interpolation.py
+++ b/animate/interpolation.py
@@ -367,7 +367,10 @@ def clement_interpolant(source, target_space=None, boundary=False):
         instructions = {
             0: "t[i] = t[i] + v[0] * s[0]",
             1: "t[i, j] = t[i, j] + v[0] * s[0, j]",
-            2: f"t[i, {dim} * j + k] = t[i, {dim} * j + k] + v[0] * s[0, {dim} * j + k]",
+            2: (
+                f"t[i, {dim} * j + k] = "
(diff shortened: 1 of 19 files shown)


animate/interpolation.py
```diff
@@ -367,7 +367,10 @@ def clement_interpolant(source, target_space=None, boundary=False):
         instructions = {
             0: "t[i] = t[i] + v[0] * s[0]",
             1: "t[i, j] = t[i, j] + v[0] * s[0, j]",
-            2: f"t[i, {dim} * j + k] = t[i, {dim} * j + k] + v[0] * s[0, {dim} * j + k]",
+            2: (
+                f"t[i, {dim} * j + k] = "
+                f"t[i, {dim} * j + k] + v[0] * s[0, {dim} * j + k]"
```
ddundo, line 372: You always put the space at the beginning of the second line and I think you like to be consistent :)
```suggestion
                f"t[i, {dim} * j + k] ="
                f" t[i, {dim} * j + k] + v[0] * s[0, {dim} * j + k]"
```

ddundo (approved):
Just one ultra minor thing but which I think you might care about :)

Issue #136, opened 2024-05-29
Parallel test hang


jwallwork23:
This week's recurring test suite failed - looks like it hung on one of the parallel tests and timed out after almost 6 hours.

```
test/test_adapt.py::test_adapt_parallel_3d_np3[False]
```

https://github.com/mesh-adaptation/animate/actions/runs/9239789581/job/25419352820

ddundo:
It looks like https://github.com/mesh-adaptation/animate/actions/runs/9282388099/job/25540301333 was hanging again on the same test so I cancelled it.

jwallwork23:
More recently we have
```
*** buffer overflow detected ***: terminated

 Unexpected error:  *** potential lack of memory.
```

Looks like there's a memory leak.

ddundo:
Another hanging test was removed in #175: `test/test_adapt.py::test_no_adapt_parallel[ParMmg]`

ddundo:
See the discussion in https://github.com/firedrakeproject/mpi-pytest/issues/12 for tips on debugging parallel tests.

stephankramer:
Right, so the first failing test `test/test_adapt.py::test_adapt_parallel_np2[mmg3d]` (i.e. 3D adaptivitiy using mmg3d with serialise=True). It seems to run into a buffer overflow in scotch called from mmg3d on one of the two processes (Note this is just serial mmg3d which seems to be calling scotch to do some renumbering, so nothing to do with repartitioning). Backtrace:
```
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007fe5a2a7c27e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007fe5a2a5f8ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007fe5a2a607b6 in __libc_message_impl (fmt=fmt@entry=0x7fe5a2c05765 "*** %s ***: terminated\n")
    at ../sysdeps/posix/libc_fatal.c:134
[...]

PR #175, opened 2025-03-01
Updates due to Firedrake's new installation process

ddundo:
Related to https://github.com/mesh-adaptation/docs/pull/114.

Sadly we now get another hanging parallel test: `test_no_adapt_parallel[Parmmg]`. I decided to remove it for now and will comment on #136 if you agree.

Edit: Parallel tests with 3 processes works (see comments below).

~~It seems like we can't run parallel tests with more than 2 processes anymore in the CI. Perhaps there's a good way to do it (ubuntu runners should support up to 4), but this isn't really a problem for us since we only have one tests with `nprocs = 3` (see comment below). I saw in https://github.com/thetisproject/thetis/pull/392 that they also don't do it; i.e., they set~~
```
    env:
      # Make sure that tests with >2 processes are not silently skipped
      PYTEST_MPI_MAX_NPROCS: 2
```
~~in their workflow. This causes the CI to fail, though. So I didn't want to include that in our workflow since, again, we only have one such test.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ lint:
 # `mpiexec -n N ... parallel[N]` only runs tests with @pytest.mark.parallel(nprocs=N)
 test: lint
 	@echo "Running all tests..."
-	@python3 -m pytest -v --durations=20 -k "not parallel" test
+	@python3 -m pytest -v --durations=20 -k "parallel[1] or not parallel" test
 	@mpiexec -n 2 python3 -m pytest -v -m parallel[2] test
 	@mpiexec -n 3 python3 -m pytest -v -m parallel[3] test
 	@echo "Done."
@@ -31,7 +31,7 @@ test: lint
 coverage:
 	@echo "Generating coverage report..."
 	@python3 -m coverage erase
-	@python3 -m coverage run --parallel-mode --source=animate -m pytest -v -k "not parallel" test
+	@python3 -m coverage run --parallel-mode --source=animate -m pytest -v -k "parallel[1] or not parallel" test
 	@mpiexec -n 2 python3 -m coverage run --parallel-mode --source=animate -m pytest -v -m parallel[2] test
 	@mpiexec -n 3 python3 -m coverage run --parallel-mode --source=animate -m pytest -v -m parallel[3] test
 	@python3 -m coverage combine
--- a/Makefile
+++ b/Makefile
@@ -22,13 +22,17 @@ lint:
 
 test: lint
 	@echo "Running all tests..."
-	@python3 -m pytest -v --durations=20 test
+	@python3 -m pytest -v --durations=20 -k "not parallel" test
+	@mpiexec -n 2 python3 -m pytest -v -m parallel[2] test
+	@mpiexec -n 3 python3 -m pytest -v -m parallel[3] test
 	@echo "Done."
 
 coverage:
 	@echo "Generating coverage report..."
 	@python3 -m coverage erase
-	@python3 -m coverage run --source=animate -m pytest -v test
+	@python3 -m coverage run --source=animate -m pytest -v -k "not parallel" test
+	@mpiexec -n 2 python3 -m coverage run --parallel-mode --source=animate -m pytest -v -m parallel[2] test
+	@mpiexec -n 3 python3 -m coverage run --parallel-mode --source=animate -m pytest -v -m parallel[3] test
 	@python3 -m coverage html
 	@echo "Done."
 
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ lint:
 	@ruff check
 	@echo "PASS"
 
+# `mpiexec -n N ... parallel[N]` only runs tests with @pytest.mark.parallel(nprocs=N)
 test: lint
 	@echo "Running all tests..."
 	@python3 -m pytest -v --durations=20 -k "not parallel" test
@@ -30,9 +31,11 @@ test: lint
 coverage:
 	@echo "Generating coverage report..."
 	@python3 -m coverage erase
-	@python3 -m coverage run --source=animate -m pytest -v -k "not parallel" test
+	@python3 -m coverage run --parallel-mode --source=animate -m pytest -v -k "not parallel" test
 	@mpiexec -n 2 python3 -m coverage run --parallel-mode --source=animate -m pytest -v -m parallel[2] test
 	@mpiexec -n 3 python3 -m coverage run --parallel-mode --source=animate -m pytest -v -m parallel[3] test
+	@python3 -m coverage combine
+	@python3 -m coverage report -m
 	@python3 -m coverage html
 	@echo "Done."
 
--- a/test/test_adapt.py
+++ b/test/test_adapt.py
@@ -3,7 +3,7 @@ Unit tests for invoking mesh adaptation tools Mmg2d, Mmg3d, and ParMmg.
 """
 
 import importlib.util
-from pathlib import Path
+import os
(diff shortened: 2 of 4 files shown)


ddundo:
The CI fails due to `mpiexec -n 3 python3 -m pytest -v -m parallel[3] test` raising the following:

```
There are not enough slots available in the system to satisfy the 3
slots that were requested by the application:

  python3

Either request fewer slots for your application, or make more slots
available for use.

A "slot" is the Open MPI term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which Open MPI processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, Open MPI defaults to the number of processor cores

In all the above cases, if you want Open MPI to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --oversubscribe option to ignore the
number of available slots when deciding the number of processes to
launch.
```

Do we even need to run `mpiexec -n 3 python3 -m pytest -v -m parallel[3] test`? This only tests this:
```
@pytest.mark.parallel(nprocs=3)
@pytest.mark.parametrize(
    "dim,serialise",
    [(2, True)],  # [(2, True), (3, True), (3, False)], # FIXME: hang (#136)
    ids=["mmg2d"],  # ["mmg2d", "mmg3d, ParMmg"],
)
def test_adapt_parallel_np3(dim, serialise):
    """
    Test that we can successfully invoke [Par]Mmg with 3 MPI processes and that it
    changes the DoF count.
    """
    assert COMM_WORLD.size == 3
    test_adapt(dim, serialise=serialise)
```
and we already have the same test with `nprocs=2`. So maybe we can just drop this test with `nprocs=3`?

test/test_adapt.py
```diff
@@ -26,9 +27,10 @@ def load_mesh(fname):
     :return: the mesh
     :rtype: :class:`firedrake.mesh.MeshGeometry`
     """
-    venv = os.environ.get("VIRTUAL_ENV")
-    mesh_dir = os.path.join(venv, "src", "firedrake", "tests", "firedrake", "meshes")
-    return Mesh(os.path.join(mesh_dir, fname + ".msh"))
+    firedrake_spec = importlib.util.find_spec("firedrake")
+    firedrake_basedir = Path(firedrake_spec.origin).parent.parent
+    mesh_dir = firedrake_basedir / "tests" / "firedrake" / "meshes"
+    return Mesh(str(mesh_dir / f"{fname}.msh"))
```
jwallwork23, line 33: Is there a reason you switched from `os.path` to `pathlib` other than personal preference? Just out of interest as I've always just used `os.path` for handling paths.

jwallwork23:
> Sadly we now get another hanging parallel test: `test_no_adapt_parallel[Parmmg]`. I decided to remove it for now and will comment on #136 if you agree.

Alas. Sure, we can do that but fixing this is becoming very much of a priority. We should address it soon after merging this.
 
> It seems like we can't run parallel tests with more than 2 processes anymore in the CI. Perhaps there's a good way to do it (ubuntu runners should support up to 4), but this isn't really a problem for us since we only have one tests with `nprocs = 3` (see comment below). I saw in [thetisproject/thetis#392](https://github.com/thetisproject/thetis/pull/392) that they also don't do it; i.e., they set
> 
> ```
>     env:
>       # Make sure that tests with >2 processes are not silently skipped
>       PYTEST_MPI_MAX_NPROCS: 2
> ```
> 
> in their workflow. This causes the CI to fail, though. So I didn't want to include that in our workflow since, again, we only have one such test.

Ah okay maybe so. In the case of Animate it would be useful to be able to use more than 2 processes because the ParMmg cases become much more complex with 3 or more. But if it's not possible then maybe we'll just have to run such tests locally.

jwallwork23 (approved):
Thanks again @ddundo.

stephankramer:
Yes I agree, I think it would be important to keep tests with nproc>2 as they tend to hit different corner cases w.r.t. local and halo numbering, ownership etc. (2 procs is the easy case). Thetis doesn't bother with too many parallel tests as it relies on firedrake having that all implemented correctly (and testing it). However, here we are also testing whether the firedrake->dmplex->petsc->mmg interface with all its fragile renumberings continues to work correctly.

It's a little unclear whether the 4 cpus that github promises for runners for public repos, are actual cores or hyperthread cpus. The fact that openmpi thinks there are only 2 suggests the latter. So it's worth trying to run with ` --use-hwthread-cpus` as suggested in that error message.

test/test_adapt.py
ddundo: Ah sorry, I was trying out a way to do all of these steps using only `pathlib` but in the end I had to use `importlib`. I'll revert to `os` to be consistent with the rest of the code.

Makefile
```diff
@@ -20,15 +20,22 @@ lint:
...
+	@mpiexec -n 3 python3 -m pytest -v -m parallel[3] test
 	@echo "Done."
 
 coverage:
 	@echo "Generating coverage report..."
 	@python3 -m coverage erase
-	@python3 -m coverage run --source=animate -m pytest -v test
+	@python3 -m coverage run --parallel-mode --source=animate -m pytest -v -k "parallel[1] or not parallel" test
+	@mpiexec -n 2 python3 -m coverage run --parallel-mode --source=animate -m pytest -v -m parallel[2] test
+	@mpiexec -n 3 python3 -m coverage run --parallel-mode --source=animate -m pytest -v -m parallel[3] test
```
stephankramer, line 36: Should the final `-m`s in these two lines be `-k`s ?

test/test_adapt.py
```diff
@@ -90,7 +92,8 @@ def test_no_adapt(dim, serialise):
 
 @pytest.mark.parallel(nprocs=2)
 @pytest.mark.parametrize(
-    "dim,serialise", [(3, True), (3, False)], ids=["mmg3d", "ParMmg"]
+    # "dim,serialise", [(3, True), (3, False)], ids=["mmg3d", "ParMmg"]  # Hangs (#136)
+    "dim,serialise", [(3, True),], ids=["mmg3d",]
```
stephankramer, line 96: Do the others not work?

Makefile
ddundo: In all three lines, it seems to me like `-k` and `-m` are interchangeable here; i.e., I do not see a difference when I run both. I think this is expected behaviour of `firedrakeproject/mpi-pytest`.

Should I change it to `-k` here and in other repos?

test/test_adapt.py
ddundo: The `ParMmg` one hangs unfortunately (see https://github.com/mesh-adaptation/animate/actions/runs/13608116705/job/38042145519#step:6:358 ). As agreed with Joe, I removed it here for now and agreed that we should address all hanging tests asap (#136).

Makefile
stephankramer: TIL about -m for pytest

test/test_adapt.py
stephankramer: Ah ok yes sounds fine

stephankramer (approved):

jwallwork23:
Can we merge this? I got myself confused with why the CI wasn't working (#177).

PR #182, opened 2025-05-08
Fix failure to import with prefix-installed petsc

### LINKED ISSUE #174: fail to import with prefix-installed petsc
The following line in animate/quality.py
```
PETSC_DIR, PETSC_ARCH = get_petsc_dir()
```
breaks _on import of animate_ when using a prefix-installed petsc. For prefix installs PETSC_ARCH is empty and get_petsc_dir() only returns a single directory. If the include directory we are after is $PETSC_DIR/$PETSC_ARCH/include/eigen3 we should be using `"%/include/eigen3" %get_petsc_dir()[-1]` I think.

Related to this, do we _have_ to use Eigen for this? Note that it is no longer included in a standard firedrake petsc configuration, so it would be nice if this could be avoided.

stephankramer:
Closes #174.

Rationale: for a local petsc build, get_petsc_dir() returns two directories: 1) $PETSC_DIR 2) $PETSC_DIR/$PETSC_ARCH. For prefix installed petsc where PETSC_ARCH is not set, it only returns one directory, namely $PETSC_DIR. For the purpose of finding the eigen header files we always want to use the last of the directories returned by get_petsc_dir().

### DIFF
--- a/animate/quality.py
+++ b/animate/quality.py
@@ -11,8 +11,8 @@ from firedrake.petsc import PETSc
 from pyop2 import op2
 from pyop2.utils import get_petsc_dir
 
-PETSC_DIR, PETSC_ARCH = get_petsc_dir()
-include_dir = ["%s/include/eigen3" % PETSC_ARCH]
+petsc_dirs = get_petsc_dir()
+include_dir = ["%s/include/eigen3" % petsc_dirs[-1]]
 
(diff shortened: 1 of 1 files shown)


jwallwork23 (approved):
Ah okay, thanks for this @stephankramer.

PR #184, opened 2025-05-15
Reinstate mmg3d in serial and serialised parallel

stephankramer:
Partially addresses #136
CI container now uses ptscotch v7.0.8 (see https://github.com/mesh-adaptation/docs/pull/140) which appears to fix previously failing mmg3d tests.


jwallwork23 (approved):
Great stuff! Thanks @stephankramer

PR #187, opened 2025-05-18
Fix adjoint interpolation

### LINKED ISSUE #113: Adjoint interpolation operator is incorrect

jwallwork23:
Closes #113.

Fixes adjoint interpolation by making use of the `adjoint=True` kwarg for the `Function.interpolate` method and by modifying `function2cofunction` and `cofunction2function` so that they accept rvalues to allow them to modify in-place.

Re-enabled the adjoint transfer unit tests for the cases with the same function spaces. If the forward transfer operator is the identity then so is the adjoint.

### DIFF
--- a/animate/interpolation.py
+++ b/animate/interpolation.py
@@ -232,27 +232,25 @@ def _transfer_adjoint(target_b, source_b, transfer_method, **kwargs):
     bounded = is_project and kwargs.pop("bounded", False)
 
     # Map to Functions to apply the adjoint transfer
-    if not isinstance(target_b, firedrake.Function):
-        target_b = cofunction2function(target_b)
-    if not isinstance(source_b, firedrake.Function):
-        source_b = cofunction2function(source_b)
+    target_b_func = cofunction2function(target_b)
+    source_b_func = cofunction2function(source_b)
 
-    Vt = target_b.function_space()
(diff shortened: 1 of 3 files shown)


stephankramer:
I think you already said something that was missing when we looked at it in the meeting, but I didn't quite catch it - so apologies if this is duplicate:

It looks like `_transfer_adjoint()` is not actually called with `transfer_method='interpolation'` in any of the tests (I put a `assert not is_project` on the 2nd line and its still passes all tests).

It doesn't look like Function.interpolate() has an `adjoint=True` flag, that only exists for the Interpolator methods.

I haven't really looked into to maths side of this, but I'm wondering whether the reason we we were getting the wrong answer previously is because we are first transferring the adjoint solution from a cofunction to a function (which is mathematically a bit dodgy). If you `cofunction1.interpolate(confunction2)` I _think_ it does the right thing , which is that for the resulting `confunction1` we have `confunction1(function1.interpolate(function2)) = confunction2(function2)` - no `adjoint=True` flag needed.

jwallwork23:
Ah thanks very much for this, it's enlightening. Seems like we need to rework things a bit.

PR #188, opened 2025-05-18
Implement restrict_anisotropy_first

### LINKED ISSUE #100: Implement `dm_plex_metric_restrict_anisotropy_first`
This should be easy enough to implement (just tweak the arguments to the `enforce_spd` call in the definition of `normalise`). However, I'd prefer to address it in a separate issue rather than here, considering the impact with some tests.

_Originally posted by @jwallwork23 in https://github.com/pyroteus/animate/pull/99#discussion_r1539967471_
            
Note that the parameter already exists on the PETSc side but is unused by Animate because it implements normalisation itself rather than with PETSc.

### LINKED ISSUE #178: Normalising for target complexity is unreliable
Because of the ordering of rescaling and imposing min and max sizes and aspect ratio, the target complexity is not always achieved. I understand this is often unavoidable w.r.t. h_min and h_max, so there you have to sort of rely that you're not hitting the min and max bounds in too many places to get a complexity that's reasonably close to target. Another issue that's a bit harder to control is hitting the maximum aspect ratio. As an example where there's only variation in a single direction, take the 2D function f(x, y) = 100 x^2 with Hessian H = [[100,0],[0,0]] with the following very reasonable parameters:
```
from firedrake import *
from animate import *

mesh = UnitSquareMesh(10, 10)
TV = TensorFunctionSpace(mesh, "CG", 1)
H = RiemannianMetric(TV)
H.interpolate(as_tensor([[100,0],[0,0]]))
H.set_parameters(
        {'dm_plex_metric_h_min': 1e-5,
         'dm_plex_metric_h_max': .2,
         'dm_plex_metric_a_max': 10,
         'dm_plex_metric_target_complexity': 1000}
        )

H.normalise()
print(H.complexity())
print(H.dat.data[0])
```
This produces a complexity of 1e9. The reason for that is that inside normalize we first call `self.enforce_spd(restrict_sizes=False, restrict_anisotropy=False)` which simply applies an arbitrary aspect ratio of 1e15 and thus we end up with [[100,0],[0,1e-28]] - keeping the largest eigenvalue at 100 (i.e. smallest edge length .1) but changing the smallest eigenvalue to 1e-28 (edge length 1e14). This results in a complexity of 1e-13 so then it scales everything up by a factor 1e16 to achieve the target complexity of a 1000, which means dividing all edge lengths by 1e8, giving [[1e18, 0], [0, 1e-12]], and min/max edge lengths of 1e-9 and 1e6. This then hits h_min, and with an aspect ratio of 10 we end up with 1e-5 in the x direction and 1e-4 in the y direction, giving the complexity of 1e9.

Given that the normalisation does not change the aspect ratio, would it not make more sense to always apply the aspect ratio bound first?

jwallwork23:
Closes #100.
Closes #178

Allows the user to restrict anisotropy before normalising the metric. I couldn't think of a useful test for this functionality. Any thoughts?

### DIFF
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -75,6 +75,7 @@ class RiemannianMetric(ffunc.Function):
         if isinstance(function_space, fmesh.MeshGeometry):
             function_space = ffs.TensorFunctionSpace(function_space, "CG", 1)
         self._metric_parameters = {}
+        self._restrict_anisotropy_first = False
         metric_parameters = kwargs.pop("metric_parameters", {})
         super().__init__(function_space, *args, **kwargs)
 
@@ -226,8 +227,8 @@ class RiemannianMetric(ffunc.Function):
             https://www.mmgtools.org/mmg-remesher-try-mmg/mmg-remesher-options/mmg-remesher-option-v.
         * `isotropic`: Optimisation for isotropic metrics. (Currently unsupported.)
         * `uniform`: Optimisation for uniform metrics. (Currently unsupported.)
-        * `restrict_anisotropy_first`: Specify that anisotropy should be restricted
-            before normalisation? (Currently unsupported.)
+        * `restrict_anisotropy_first`: Boolean flag specifying that anisotropy should be
+            restricted before normalisation when True. Default: False.
 
         :kwarg metric_parameters: parameters as above
         :type metric_parameters: :class:`dict` with :class:`str` keys and value which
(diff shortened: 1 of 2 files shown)


jwallwork23:
Thanks @stephankramer. I've added some tests so this is ready for review now.

stephankramer (approved):
Looks great, thanks

Issue #197, opened 2025-07-16
`test_adapt.py::test_adapt[mmg3d]` fails in CI


ddundo:
See https://github.com/mesh-adaptation/animate/actions/runs/16244614135/job/46127219579.

It doesn't give a useful error message, only `Error: Process completed with exit code 1.`. From [pytest docs](https://docs.pytest.org/en/7.1.x/reference/exit-codes.html): `Exit code 1 Tests were collected and run but some of the tests failed`

PR #201, opened 2025-09-02
Avoid use of `.vector()`

### LINKED ISSUE #200: `vector` method no longer supported
Recent CI jobs fail because the `vector` method for `Function`s appears to have been removed in https://github.com/firedrakeproject/firedrake/pull/4499.

See also https://github.com/mesh-adaptation/goalie/issues/316.

jwallwork23:
Closes #200.

This PR replaces all uses of `Function.vector().gather()` with `.dat.global_data` to adopt new API.

It also drops use of `MeshGeometry.init()`, which no longer exists.

This PR also temporarily turns off other broken 3D tests reported in #197.

### DIFF
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -496,13 +496,13 @@ class RiemannianMetric(ffunc.Function):
         a_max = interp(self._variable_parameters["dm_plex_metric_a_max"])
 
         # Check minimal h_min value is positive and smaller than minimal h_max value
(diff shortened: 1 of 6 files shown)


stephankramer:
@jwallwork23, as discussed yesterday: I've just opened an alternative to this that also gets rid of the (implicit) all-to-all-gather: #202
It copies the disabling of the 3D tests, and the mesh init from here.

PR #202, opened 2025-09-19
Avoid use of .vector().gather()

### LINKED ISSUE #200: `vector` method no longer supported

stephankramer:
Closes #200

Alternative for #201.
Removes usage of .vector()
Taking this opportunity to also eliminate .vector().gather() (which now would be .dat.global_data) as all-to-all gathers of function data should generally be avoided. All places where this was used were actually doing a reduction operation straight after, so now replaced
 with just an MPI allreduce. Introducing utility functions
function_data_min/max/sum for this as it gets a little wieldy.

Copies the other parts of #201:
It also drops use of MeshGeometry.init(), which no longer exists.

This PR also temporarily turns off other broken 3D tests reported in #197.

### DIFF
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -24,6 +24,7 @@ from .recovery import (
     recover_gradient_l2,
     recover_hessian_clement,
 )
+from .utility import function_data_min, function_data_sum
 
 __all__ = ["RiemannianMetric", "determine_metric_complexity", "intersect_on_boundary"]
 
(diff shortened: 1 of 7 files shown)


jwallwork23 (approved):
This is great, thanks @stephankramer! I'll close #201 as this is indeed a much better approach.

PR #204, opened 2025-10-08
Include cxx files in installed Python package

angus-g:
Maybe this is naive of me, but this should enable non-editable installs to succeed as well?


stephankramer (approved):
Thanks @angus-g . That seems indeed what's required (tested locally)

jwallwork23:
Merging now that the CI is back up and running.

PR #205, opened 2025-10-10
Fix some of the tests

### LINKED ISSUE #197: `test_adapt.py::test_adapt[mmg3d]` fails in CI

joewallwork:
Closes #197.

I just rebuilt Firedrake locally and found some tests are now working.

However, I noticed that the tests that use meshes that're in `firedrake/tests/firedrake/meshes` don't always pick the path up correctly. The current setup looks inside `site-packages` but that's not where my Firedrake build is. I've copied the mesh files over temporarily while we figure out a solution.

### DIFF
--- a/animate/metric.py
+++ b/animate/metric.py
@@ -302,7 +302,7 @@ class RiemannianMetric(ffunc.Function):
         consistent.
         """
         entity_dofs = np.zeros(self._tdim + 1, dtype=np.int32)
-        entity_dofs[0] = self._mesh.geometric_dimension()
+        entity_dofs[0] = self._mesh.geometric_dimension
         coord_section = self._mesh.create_section(entity_dofs)[0]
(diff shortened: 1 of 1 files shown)


jwallwork23:
Okay this fixes various things but there are low-level things we'll have to handle to follow up. If we merge this PR then my changes in https://github.com/mesh-adaptation/movement/pull/166 should pass.

jwallwork23:
Same here - gonna merge it to get the ball rolling.

PR #208, opened 2025-11-14
Add codeowners

joewallwork:
This PR adds a `CODEOWNERS` file so that the codeowners mentioned will automatically be added as reviewers for any PR that gets opened. This is useful because external contributors can't request reviewers.

I also took the opportunity to update the authors and maintainers lists.


stephankramer (approved):
Clever, all makes sense

PR #218, opened 2026-01-16
Separate `main` and `release` workflows

joewallwork:
See https://github.com/mesh-adaptation/docs/issues/156.

This PR adds a separate test suite workflow for the new `release` branch. That test suite not only runs on Animate's `release` branch, but also with the `release` version of Firedrake.


stephankramer (approved):
I just reran the tests to confirm that it's now actually failing on release :-)
which I guess should be fixed in a PR into release afterwards
But yes, good to go

PR #226, opened 2026-02-20
Move `to_petsc_local_numbering` into Animate

### LINKED ISSUE #225: Move `to_petsc_local_numbering` into Animate?
This would require adding a Cython module but it'd give us better control over the interface and would help to remove unused functionality from Firedrake.

joewallwork:
Closes #225.

Once this is merged, I'll open a PR to remove `to_petsc_local_numbering` from Firedrake.

### DIFF
--- a/test/test_numbering.py
+++ b/test/test_numbering.py
@@ -7,15 +7,33 @@ from animate.cython.numbering import to_petsc_local_numbering
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


stephankramer:
This works indeed for me locally. I think this should work with https://github.com/mesh-adaptation/docs/pull/173

PR #237, opened 2026-04-02
Use petsctools.get_petsc_dir instead of pyop2.utils

### LINKED ISSUE #492: Render CI fails: animate incompatible with latest firedrake (get_petsc_dir removed from pyop2)
The render workflow is failing because the animate package (installed from main) tries to import get_petsc_dir from pyop2.utils, which was removed in firedrake PR firedrakeproject/firedrake#5002. That refactor moved the function into the standalone petsctools package.

The failing demo is adaptive_base_case, which imports from animate. The traceback is:

    from animate import RiemannianMetric, adapt
    -> animate/quality.py line 12: from pyop2.utils import get_petsc_dir
    -> ImportError: cannot import name 'get_petsc_dir' from pyop2.utils

Interestingly, animate own setup.py already uses petsctools.get_petsc_dir(), so quality.py is the only file that was missed.

Opening a fix upstream on mesh-adaptation/animate.

sghelichkhani:
Firedrake PR firedrakeproject/firedrake#5002 removed get_petsc_dir from pyop2.utils and moved it into the standalone petsctools package. The setup.py here already uses petsctools, but quality.py was still importing from pyop2.utils, which now breaks at import time.

This is a one-line fix, changing the import in quality.py from pyop2.utils to petsctools.

Fixes g-adopt/g-adopt#492

### DIFF
--- a/animate/quality.py
+++ b/animate/quality.py
@@ -8,10 +8,10 @@ import firedrake
 import ufl
 from firedrake.__future__ import interpolate
 from firedrake.petsc import PETSc
+from petsctools import get_petsc_dirs
 from pyop2 import op2
-from pyop2.utils import get_petsc_dir
 
-petsc_dirs = get_petsc_dir()
+petsc_dirs = get_petsc_dirs()
 include_dir = ["%s/include/eigen3" % petsc_dirs[-1]]
 
 __all__ = ["QualityMeasure"]
--- a/test/test_checkpoint.py
+++ b/test/test_checkpoint.py
@@ -7,7 +7,6 @@ import numpy as np
 import ufl
 from firedrake.function import Function
 from firedrake.functionspace import FunctionSpace, TensorFunctionSpace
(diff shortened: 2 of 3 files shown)


animate/quality.py
```diff
@@ -9,7 +9,7 @@
 from firedrake.__future__ import interpolate
 from firedrake.petsc import PETSc
 from pyop2 import op2
-from pyop2.utils import get_petsc_dir
+from petsctools import get_petsc_dir
```
stephankramer:
```suggestion
from petsctools import get_petsc_dir
from pyop2 import op2
```

stephankramer (changes requested):
Thanks Sia, these two lines need swapping to pass `make lint`

animate/quality.py
sghelichkhani: Done! Thanks Stephan!

stephankramer (approved):
Got there in the end!

PR #239, opened 2026-05-21
Merge main into release

stephankramer:
This PR undoes #230 completely because it consists of a number of commits that are already on main (see https://github.com/mesh-adaptation/animate/pull/230/commits) combined with a number of commits to undo changes on main (because the starting point for release wasn't actually compatible with then Firedrake release) - and these commits were all squashed into one. Meaning that after merging main into release, we will have the same changes twice in history, and also the undone changes would need to be re-done.

By adding an explicit revert commit, the combined effect of 68242c8855ed74b9198b85c28a6d110ce03645a4 (the squashed commit of #230) and the revert commit is zero and thus we can cleanly merge main into release - at which point they are the same - which I think (hope!) is compatible with Firedrake release at the moment.

The only alternative I can see is if we discontinue the history of release - which is fine, we just need to warn people to delete their local release branch - so let me know if you prefer that.


stephankramer:
Oh I guess I first need to fix the release mesh-adaptation container build....

stephankramer:
This appears to work now in the updated CI release container.

@joewallwork : this is now current release, which contains two commits 68242c8855ed74b9198b85c28a6d110ce03645a4 and 6c141066c1c2610b979e6e61d9749c150c4be788 that are not on main, followed by two revert commits, followed by a merging in of main. We could also simply reset release to main (i.e. discontinue release history) if you prefer. Moving forward, we should probably decide on a policy for the maintenance of release. If we want to maintain a continuous history for both release and main, we need to be quite careful how we handle both, i.e. no cherry-picking or rebasing of commits that are already in the other branch, a hot-fix needs to go in release first and then be merged into main such that release is always an ancestor of main, etc. Discontinuing release - i.e. resetting it whenever we want to sync it with main, probably after each firedrake release - makes life a lot easier but you have to warn people to not "git pull" on their local release checkout. FWIW g-adopt and probably Thetis are going for the discontinuous approach.

joewallwork (approved):
Hi @stephankramer, apologies that it took me so long to look at this. I'm happy to go with what you have here for now and to move to a discontinuous approach if needed, as it sounds like that is the recommended approach.

PR #243, opened 2026-07-23
Fix workflows to deal with PRs to release

stephankramer:
Current test_suite_main triggers on every PR and runs the *main* test suite on the PR branch even when the PR targets release.

When a PR targets release it should run the PR branch in the release CI workflow (i.e. test it against firedrake release). Also after the merge into release (or any other change to release), we should run the *release* workflow on it to check everything is still solid.

I've also removed the path filters because I think they are a bit unsafe, it's very easy to overlook a file that might influence the correct running of the tests - and since the documentation lives elsewhere I think there are very few commits that do not need to be tested.

### DIFF
--- a/.github/workflows/test_suite_main.yml
+++ b/.github/workflows/test_suite_main.yml
@@ -5,21 +5,11 @@ on:
   push:
     branches:
       - main
-    paths:
-      - '**.py'
-      - '**.cxx'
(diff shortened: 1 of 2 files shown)


joewallwork (approved):
Ah good spot!

You are probably right - the path filters don't achieve that much for this repo.

PR #244, opened 2026-07-28
Remove from __future__ import interpolation

stephankramer:
No longer exists in firedrake main. The future is now!

Changed some firedrake.assemble(firedrake.interpolate(x, V)) to firedrake.Function(V).interpolate(x) to reduce line lengths.


joewallwork (approved):
Nice. Thanks @stephankramer

